PR #15, opened 2017-10-02
Include Photon performance dashboard

armenzg:


armenzg:
@digitarald Hi Harald, would you mind having a quick look (FYI this is still WIP) but I believe I'm going in the right track. I'm now going to look at the code that consumes and displays evolutions.

I also removed some mission-control code as I think I was getting 'MissionControl' is defined but never used. (no-unused-vars' in the linter.

Thanks in advance!

armenzg:
The work has moved to another PR.

Issue #8, opened 2017-09-07
investigate adding photon dashboard


stuartphilp:
preferably directly pull in, but at least link to the photo dashboard at: http://astithas.com/perm/photon-perf/

this should be in spot #3, just before page load times

armenzg:
I will be looking into this.

armenzg:
@digitarald do you have any suggestions as to how to integrate this? (In case you have already given it some thought).

digitarald:
https://github.com/mozilla/platform-health/blob/6d5b10a3c0a72318aba269530551575f2dd9284e/src/perf.js#L395 has some code that grabs telemetry graphs over versions. It was driving https://health.graphics/quantum/pageload/render (other format: https://health.graphics/quantum/track?metric=GC_MAX_PAUSE_MS&child=content ) , which seems to be broken right now.

armenzg:
I'm hitting this:
```bash
ERROR in ./node_modules/bindings/bindings.js
Module not found: Error: Can't resolve 'fs' in '/Users/armenzg/repos/platform-health/node_modules/bindings'
 @ ./node_modules/bindings/bindings.js 6:9-22
 @ ./node_modules/hiredis/hiredis.js
 @ ./node_modules/redis-parser/lib/hiredis.js
 @ ./node_modules/redis-parser/lib/parser.js
 @ ./node_modules/redis-parser/index.js
 @ ./node_modules/redis/index.js
 @ ./node_modules/then-redis/lib/Client.js
 @ ./node_modules/then-redis/lib/index.js
 @ ./src/fetch/text.js
 @ ./src/fetch/json.js
 @ ./src/release/versions.js
 @ ./static/quantum/photon_perf.js
 @ ./static/quantum/index.js
 @ ./static/routes.js
 @ ./static/index.js
 @ multi react-hot-loader/patch webpack-hot-middleware/client?reload=true ./index.js
```

`yarn add fs` does not help.

digitarald:
Looks like your node installation is broken, `fs` comes with node.

armenzg:
I removed node_modules and installed again. At the moment I don't hit this issue anymore.

armenzg:
I believe I've figured out how to use telemetry-wrapper to plot this within React:
https://github.com/armenzg/react-telemetry-demo/commit/07f1fb1a0cf1f837364b0a493e142e780be5789b

If @digitarald finds no objections with this approach I will move forward with it.

armenzg:
[...]

PR #18, opened 2017-10-11
Issue #8 - Add Photon Perf dashboard into Quantum page

armenzg:
I need to do some CSS changes but this shows the 4 graphs being plotted.

### DIFF
--- a/src/perf/tmo-wrapper.js
+++ b/src/perf/tmo-wrapper.js
@@ -1,0 +1,123 @@
+// Module with partial logic from
+// https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/wrapper/telemetry-wrapper.js
+// This module focuses on data manipulations rather than data plotting
+import { stringify } from 'query-string';
+
+import Telemetry from './telemetry-node';
+import { getEvolution } from './tmo';
+import getVersions from '../release/versions';
+import TELEMETRY_CONFIG from './telemetry_config';
+
+// This function is a copy/paste from telemetry-wrapper
+const createTMOLinkForParams = (params) => {
+  const url = `https://telemetry.mozilla.org/new-pipeline/${params.evoVersions ? 'evo' : 'dist'}.html#!`;
+
+  let queryParams = stringify({
+    max_channel_version: `${params.channel}%252F${params.version}`,
+    sanitize: `${params.sanitize ? 1 : 0}`,
+    trim: `${params.trim ? 1 : 0}`,
+    use_submission_date: `${params.useSubmissionDate ? 1 : 0}`,
+    measure: `${params.metric}`,
+  });
+  if (params.evoVersions) {
+    queryParams += stringify({
+      min_channel_version: `${params.channel}%252F${params.version - params.evoVersions + 1}`,
+    });
+  }
+
+  if (params.filters) {
+    ['os', 'application', 'e10sEnabled', 'child'].forEach(key =>
+      queryParams += stringify({
+        key: params.filters[key],
+      }),
+    );
+  }
+
+  // Special case: dashgen defaults to not filtering anything
+  // but the dashes default to filtering to only Firefox Desktop
+  if (!(params.filters && 'application' in params.filters)) {
+    // This isn't all products, but it's close enough
+    queryParams += stringify({
+      product: 'Firefox!Fennec',
+    });
+  }
+
+  if (params.compare) {
+    queryParams += stringify({
+      compare: `${params.compare}`,
+    });
+  }
+
+  return `${url}?${queryParams}`;
+};
+
+// Reduced function from telemetry-wrapper
+const setDefaultParams = (params) => {
+  params.useSubmissionDate = params.useSubmissionDate || false;
+  params.percentile = params.percentile || 50;
+  return params;
+};
+
+// Partial logic from evoTime in telemetry-wrapper
+const prepareDataForGraph = (params, evolutions, versions) => {
+  const dateses = evolutions.map(evo => evo.dates());
+  const legendLabels = versions.map(v => v.toString());
+  let percentileLabel = 'median';
+  if (params.percentile !== 50) {
+    percentileLabel = `${params.percentile}th percentile'`;
+  }
+  const yLabel = percentileLabel;
+  const description = evolutions[0].description;
+  const valueses = evolutions.map(evo => evo.percentiles(params.percentile));
+  const datas = dateses.map((dates, i) => dates.map((date, j) => {
+    return {
+      date: date.toISOString().substring(0, 10),
+      value: valueses[i][j],
+    };
+  }));
+
+  return {
+    datas,
+    description,
+    legendLabels,
+    params,
+    yLabel,
+  };
+};
+
+export const fetchTelemetryEvolution = async (ctx, name) => {
+  await new Promise((resolve) => {
+    Telemetry.init(resolve);
+  });
+  const channelVersions = await getVersions();
+  const params = setDefaultParams(TELEMETRY_CONFIG[name]);
+
+  const start = parseInt(channelVersions.nightly, 10);
+  const endVersion = start - params.evoVersions;
+
+  const versions = [];
+  for (let version = start; version > endVersion; version -= 1) {
+    versions.push(version);
+  }
+  const evolutionMap = await Promise.all(
+    versions.map(async (version) => {
+      const query = Object.assign({}, {
+        metric: params.metric,
+        channel: params.channel,
+        filters: params.filters,
+        useSubmissionDate: params.useSubmissionDate,
+        version: version,
+      });
+      return getEvolution(query);
+    }),
+  );
+  ctx.body = {
+    identifier: name,
+    graphData: prepareDataForGraph(
+      params,
+      // Remove evolutions that failed
+      evolutionMap.filter(e => e !== null).reverse(),
+      versions.reverse()),
+    telemetryUrl: createTMOLinkForParams(params),
+  };
+};
--- a/static/telemetry/graph.js
+++ b/static/telemetry/graph.js
@@ -1,0 +1,67 @@
+/* global fetch */
+import MG from 'metrics-graphics';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export default class TelemetryContainer extends React.Component {
+  async componentDidMount() {
+    const { id } = this.props;
+    this.fetchPlotGraph(id);
+  }
+
+  async fetchPlotGraph(id) {
+    const { graphData, telemetryUrl } = await (
+      await fetch(`/api/perf/telemetry/${id}`)).json();
+    this.graphTitleLink.setAttribute('href', telemetryUrl);
+    this.graphSubtitleEl.textContent = graphData.description;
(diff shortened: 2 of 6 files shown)


armenzg:
I removed the vendoring of telemetry-wrapper.

I still have to tackle the CSS/divs issues (see screenshot):
https://cl.ly/1Y2C1r0E330i

armenzg:
@digitarald @stuartphilp r?

Here's a screenshot: https://cl.ly/1v2E1h3O1C1A
Here's the deployment of it: https://health-quantum-dashboard.herokuapp.com/quantum

armenzg:
CC: @past (if curious)

stuartphilp:
@armenzg is the other pr still needed or is this inclusive? one change is maybe make the graphs more consistent with the others in terms of style (longer/flatter instead of square, white background, grey title bar). but other wise it looks great and the separation between releases is nice :)

digitarald:
Thinking about this some more now. Sorry about the lack of feedback the past days.

Generally, can we break out the cleanup commits into a separate pull request while this one gets refined?

I don't think we have good reasons for this patch to break the existing practices that all other code in this repo follows:
- All other data is served through the server to have fine-grained control over caching
- Data being rendered in React

tmo can take long to query, especially going back several versions; making the graphs slow to appear on TVs. Given that and the amount of dependencies injected into index.html just to draw a new set of charts this needs some refactoring.

The project also has existing server code (that might be broken right now, tbc) that queries t.m.o for historical data across versions; this should be a good starting point for creating the data source:

https://github.com/mozilla/platform-health/blob/master/src/perf.js#L395

There are few examples on how to draw a graph using SVG + d3. It definitely feels repetitive right now, maybe there can be some more convenience provided to factor out containers, axis and lines into components.

armenzg:
Can I please be made a contributor to this project? (I would like to be able to set the "reviewer" flag).

@stuartphilp I closed the other PR. As part of the new approach I will make the graphs more consistent.

@digitarald Can you please have a look at this?
https://github.com/mozilla/platform-health/pull/18/commits/15c2ce45e76a588da25df3ee138667f1d09187a8
I've moved the fetching of evolutions to the backend.

While you look at this backend change I will remove some of the <script> logic and iterate on the frontend.

digitarald:
@armenzg I have not pulled 15c2ce4 down for testing, but the approach looks correct. It might be better to have one endpoint for each telemetry probe, so each graph can pull data in parallel

armenzg:
@digitarald no need to pull changes and test them. This is still WIP.

I switched from one endpoint to four:
https://github.com/mozilla/platform-health/pull/18/commits/ff8e1ef079098b10135b6ba1828a930798eb6658#diff-da0eb620dc7b8be4a526a465b6cd7152R395

Would that meet your expectations?

I'm also trying to grab logic from telemetry-wrapper (the function evoTime) so I don't have to do manipulations of svgs.
I will be honest the svg code is something completely new to me. It will take me a bit to grasping it.
Would something going in this direction be acceptable? (see commit below) Or do you have a component that generate svg for the same kind of data? (or very similar)

NOTE: This is not yet working. More logic from telemetry-wrapper needs to be grabbed:
https://github.com/mozilla/platform-health/pull/18/commits/d98bb75c82aa0aa73678bb1759bde33e0f8e03ce#diff-6971ba22a431a2291ae4e2dc946edb82

armenzg:
@digitarald I'm making progress with the approach I devised.

Could you please check that it is acceptable for you?
https://github.com/mozilla/platform-health/pull/18/commits/2052102a68878a04bac77157cf1fc9265c0c68ad#diff-6971ba22a431a2291ae4e2dc946edb82R102

I still have to polish the call to evoTime() to get the graph to be drawn properly.

`telemetryConfigureElements` can easily be turned into a component.

`createGraphEls` and `evoTime` are modified versions of telemetry-wrapper.
I'm hoping that some of what I accomplish here can lay the grounds for the telemetry team having a reference to release telemetry-wrapper as an npm package, thus, keeping some of their logic.

past:
@armenzg can I make one last-minute change request? We should limit the browser startup measurement to the parent process only (per bug [1409099](https://bugzilla.mozilla.org/show_bug.cgi?id=1409099)). I made this tiny tweak to my dashboard:

https://github.com/past/past.github.com/commit/6c795e11c6cd869dbaafd29e8229bfc885aa9b32

Thanks!

armenzg:
Thanks @past !
I've committed that change locally.

armenzg:
@digitarald for some reason I can't find the comment you made in the last few days.
I've fixed some of the things I remember reading.

I believe you mentioned trying to reuse the Graphic component.
I think it is the one used to display these 3 charts:
https://health.graphics/crashes

Using it will have some styling problems.
The styling of the Graphic/crashes page is a black background with white fonts while the quantum page is the opposite.
I *might* be able to use the Graphic component, however, I see a decent amount of styling to make both work.

I'm still prefetching CSS for metrigs-graphics and telemetry-wrapper + copying the telemetry-node.js file from src/ to static/.

At the moment, I managed to get back to drawing something half properly:
* Screenshot: https://health.graphics/crashes
* Deployed: https://health-quantum-dashboard.herokuapp.com/quantum

armenzg:
> copying the telemetry-node.js file from src/ to static/.

I've discarded the change that did that and instead added a symlink to telemetry-node:
https://github.com/mozilla/platform-health/pull/18/commits/b04af76599e05ae381f4ce63ec232decbcea35a4

I also nodged https://github.com/mozilla/telemetry-next-node/issues/2 one more time.

armenzg:
>>  copying the telemetry-node.js file from src/ to static/.
> >
> I've discarded the change that did that and instead added a symlink to telemetry-node:
b04af76
I've spoken too soon. The symlink approach only works locally.
The build is failing again on the Heroku deployment.

This can be reproduced even on a local checkout if you run this command:
`npm run clean; webpack -p`

The error has nothing that will make you think the issue is due to having telemetry-node under src/
> ERROR in app.959ff9.js from UglifyJs
> Unexpected character '`' [../src/perf/telemetry-node.js:26,0][app.959ff9.js:104400,8]
> Child html-webpack-plugin for "index.html":
>      1 asset
>       [0] ../node_modules/html-webpack-plugin/lib/loader.js!./index.html 689 bytes {0} [built]
>        + 3 hidden modules

static/quantum/photon_perf.js
```diff
@@ -0,0 +1,108 @@
...
+  if (plot.evoVersions) {
+    rootEl.getElementsByClassName('graph-container')[0].classList.add('evo-graph-container');
+    evoTime({
+      percentile: 50,
+      useSubmissionDate: plot.useSubmissionDate,
+    }, graphEl, realEvolutions, versions);
+  }
+};
+
+const createGraphEls = id => (
```
digitarald: Check out reacts refs mechanism to get hold of elements. You would move this function and processData into render and use refs to get the DOM elements that you’ll eventually need after mounting.

armenzg:
This ready for another review.

You can see the deployed PR in here:
https://health-quantum-dashboard.herokuapp.com/quantum

What do you suggest I do about mg-active-datapoint?
I'm having trouble modifying it since I can't select the element in the inspector (since it only shows when hovering over one of the data points).
I can keep trying, however, I'm hoping for a super trick to save me time.

stuartphilp:
I agree, doesn't have to be an exact match, but having them be responsive and use the same headers should be plenty to get this ready to land

armenzg:
I've addressed the comments in my latest commit.

You can see the latest deployment in here:
https://health-quantum-dashboard.herokuapp.com/quantum

Resizing seems to work.
Headers have a more similar styling. Let me know if it needs to conform to a specific header.
Plotting is still a bit odd on Chrome. Needs further investigation.

Also, I originally intended to have some functions to be as similar as possible to the original telemetry-wrapper.js, however, I'm OK with discarding any code that is not necessary. Let me know what you prefer.

image

armenzg:
I've handled showing more versions.
feedback?

I will soon be automatically deployed in here:
https://health-quantum-dashboard.herokuapp.com/quantum

image

armenzg:
I've pushed all my changes. I've squashed them a bit in my last pust.

Would you like me to address the CSS issues on Chrome? or follow up?

I believe we should file an issue make each graph have a marker for the criteria.
We can later determine the priority for it; what do you think?

image

armenzg:
I've pushed one more change to add white space between the containers (create white gap between headers).

I've also added wrapping. Handy for small screen viewing.

The deployment should be available in the next 5 minutes:
https://health-quantum-dashboard.herokuapp.com/quantum

armenzg:
Thanks for pointing that out! I've fixed it.
image

The Travis failure does not happen locally. It seems there's a socket hang.

I've squash all commits down to two.
Please merge if satisfied.

The deployment is here:
https://health-quantum-dashboard.herokuapp.com/quantum

There's a known issue when making the window width smaller and that is that a margin-left is added to the 2nd and 4th plots.
image

The CSS that causes it is also the one that keep a space between all graphs. If you have any suggestions please let me know:
 ``` javascript
 /* On a small width we would get an odd margin*/
  .criteria-widget {
    &:not(:first-child) {
      margin-left: 0;
    }
  }
```

File: src/perf/tmo-wrapper.js [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+import Telemetry from './telemetry-node';
+import { getEvolution } from './tmo';
+import getVersions from '../release/versions';
+import TELEMETRY_CONFIG from './telemetry_config';
+
+// This function is a copy/paste from telemetry-wrapper
+const createTMOLinkForParams = (params) => {
+  const url = `https://telemetry.mozilla.org/new-pipeline/${params.evoVersions ? 'evo' : 'dist'}.html#!`;
+
+  const queryParams = [];
```
digitarald: Replace this whole block of array concatenation with an object using `qs.stringify` (example [here](https://github.com/mozilla/platform-health/blob/08b87f1983c1e808dbad5905795ad728606d198f/static/quantum/perfherder.js#L30))

src/perf/tmo-wrapper.js
```diff
@@ -0,0 +1,129 @@
...
+};
+
+export const fetchTelemetryEvolution = async (ctx, name) => {
+  await new Promise((resolve) => {
+    Telemetry.init(resolve);
+  });
+  const channelVersions = await getVersions();
+  const params = setDefaultParams(TELEMETRY_CONFIG[name]);
+
+  try {
```
digitarald: Remove try/catch, as errors thrown will result in 500 anyways.

static/telemetry/graph.js
```diff
@@ -0,0 +1,85 @@
...
+      const newDate = new Date(`${YYYY}-${MM}-${DD}`);
+      return {
+        date: newDate,
+        value,
+      };
+    }));
+  }
+
+  graphEvolutionsTimeline({ datas, params, yLabel, legendLabels }, graphEl) {
+    const newDatas = this.fixDates(datas);
```
digitarald: You can probably use `MG.convert.date`, like [here](https://github.com/mozilla/platform-health/blob/4fb7e6e03c90ac82780025d9e7eab99ad0f69b8d/static/crashes/xp.js#L10).

armenzg:
The issues on Travis are unrelated to the PR.

PR #26, opened 2017-12-20
Issue 25 - Do not restrict CSS imports to only srcDir

armenzg:


armenzg:
If Eli recommended I'm happy with it.

Issue #29, opened 2018-01-17
Modernize platform-health


armenzg:
To do:
* Separate backend from frontend
* Switch frontend to Neutrino
* Make sure it deploys correctly and keeps running tests

I assume this will take few days to over a week.

The work will happen in https://github.com/armenzg/platform-health/tree/separate

armenzg:
The work is happening in here:
https://github.com/mozilla/platform-health/compare/master...armenzg:switch_neutrino

Todo:
- [ ] font loading
- [ ] fix routing
- [ ] add scripts for proper deployments

armenzg:
@digitarald @stuartphilp @eliperelman Just so this does not come as a surprise.

I've decided to separate the backend/frontend code into their own directories and boilerplate.
I can make Heroku start both up.

This will make it easier to move the forward the frontend without breaking the backend.

This all steams from that I tried to include react-metrics-graphics (for a frontend issue) and I started getting backend *production-specific* issues (https://github.com/mozilla/platform-health/issues/21#issuecomment-353190981).

I was hoping that switching to Neutrino for both parts (backend and frontend) would make it easier to add new packages but the build complexity of webpack.config.babel.js and app.js has proven challenging.
See https://github.com/mozilla/platform-health/compare/master...armenzg:switch_neutrino

If any of you have sample project I could look at to determine project layout it would be great.

digitarald:
If the outcome is a project that is as performant, easier to maintain and has better developer ergonomics, I am for it.

armenzg:
I'm going to take a step back to evaluate if we can avoid separating the project into two.
I'm going to revisit the production speficic issue when trying to introduce react-metrics-graphics:
https://github.com/mozilla/platform-health/issues/21#issuecomment-353190981

If I can't I will separate frontend and backend into different repositories since it makes the deployment story with Heroku nicer. I will explain why.

Heroku only likes one Procfile per repository and only 1 web process, thus, requiring two different apps and pushing subtrees.

The document I followed:
http://adampaxton.com/how-to-deploy-to-multiple-heroku-apps-from-the-same-git-repository/
I've managed to deploy a frontend app with code from the frontend directory, however, that requires manual intervention. It requires pushing a subtree to one of your branches.

## Side notes with regards making the static buildpack to work:

From reading this:
https://gist.github.com/hone/24b06869b4c1eca701f9#building-assets-on-heroku

Configured the Heroku remote:
[...]

PR #32, opened 2018-02-02
Issue #29 - Refactor project to a frontend only project

armenzg:
There's some docs and test changes, however, it has the meat & bones of all changes.
There's also some CSS warnings showing up when you start the project.
Please let me know how I could tackle them. I was unable to figure it out.

This is currently deployed to:
https://firefox-health-dashboard.herokuapp.com
and it fetches data from the official backend:
https://platform-health.herokuapp.com


armenzg:
@eliperelman I've landed all remaining minor changes (e.g. docs)
No more changes are expected until I have to address your feedback.
For now, I will go and finish 

I've switched the frontend to point to my backend work:
https://firefox-health-backend.herokuapp.com/

.neutrinorc.js
```diff
@@ -0,0 +1,107 @@
+const acceptedExternalEnvs = {
+  BACKEND: 'BACKEND' in process.env ?
```
armenzg: I've been trying to use Heroku Review apps and I get the feeling that if I change the env variable "post-build" it will take no effect.
Unfortunately Heroku Review apps do not allow to do a deploy from a branch as you can with normal apps, thus, needing to push code to cause another build which would include the env variable.

Restarting the Dynos of the Review App won't build again, thus, failing to use the new variable.

Pushing empty commits does not rebuild either.

The steps to make this work:
* set the env variable pointing to a different backend than the default one
* force push a dummy change

Is there a different way to rebuild a review app?

PR #52, opened 2018-02-26
Improve look of speedometer graphs

armenzg:


armenzg:
Hi @stuartphilp I don't have as much rush to land this compared to last week's changes.

In this change I have:

- Gone from 6 graphs to 4 graphs
- Added a beta line to the first graph, thus, deprecating the beta graphs
- Brought the 'Reference hardware' closer to the graphs
- Put the speedometer graphs in two columns

Please have a look the deployment and let me know if it looks good to you:
https://platform-health-pr-52.herokuapp.com/quantum
image

Issue #43, opened 2018-02-20
Support benchmarks that have subbenchmarks


armenzg:
There's some new benchmarks that we need added that need support for subbenchmarks.
This issue focuses on adding support for it.

armenzg:
By this time next week we should be having Windows running these benchmarks on mozilla-central.
I have met with @wlach and I should a prototype to look next week.

armenzg:
I have a prototype running for linux64:
https://perf-goggles.netlify.com/#/linux64/ARES6

armenzg:
- [ ] Integrate prototype into dashboard
   - [ ] Possibly need to update react router 
   - [ ] Add routes for reaching new view (e.g. quantum/talos/<platform>/<suiteName>
   - [ ] Add graphs or link to new view
   - [ ] Ask feedback
- [ ] Follow ups
   - [ ] Make graphs look native to current theming of project
   - [ ] Bookmarkability of graphs

armenzg:
Since Windows won't be ready for few more weeks.
I'm going to change this issue to giving support for subbenchmarks while moving adding the new benchmarks to a separate issue.

Issue #45, opened 2018-02-20
The status widget highlight looks awkward


armenzg:
We should *only* have white background behind the targetStatus and not see the green/yello at the top.

![image](https://user-images.githubusercontent.com/44410/36433367-539bc030-162a-11e8-9f96-79c7daa12d42.png)

csd713:
Hi @armenzg I would like to take this up. How can reproduce this issue?

armenzg:
Hi @csd713 you will have to load this page locally
http://localhost:5000/quantum and resize the window until you see the issue happening.

Unfortunately, the speedometers graphs (section "#1 - Speedometer") are not working correctly, thus, we can't reproduce the issue I reported 13 days ago.
There's an issue on the downstream service. I'm going to remove those graphs.

There's a similar issue for one of the graphs below when text is long (see green area on 2nd graph; the smiley face is being hidden):
![image](https://user-images.githubusercontent.com/44410/36991476-c4c39d3c-2075-11e8-80ed-2fe87a5eeb88.png)

Since the original issue cannot be reproduced anymore we could close the issue if you like since the new issue is not as noticeable.

Please let me know what you prefer. I'm happy either way.

csd713:
Hello @armenzg I was able to reproduce this issue on a bigger screen (15 inch MBP).
I tried to wrap the background-color for the text, but I couldn't find a solution. I would really appreciate if could you help me with any pointer to achieve it.

armenzg:
Hi @csd713 I'm not very good at CSS so I will share some of what I discovered and hopefully that will trigger ideas to try. This is not a very important issue in case you end up deciding not to spend more time on it. It might be a good learning opportunity but don't worry too much if you decide to give up.

I used the dev tools inspector to look at the <header>, <h3> and <aside> elements.
If we disable (you can disable directly in the devtools and see the changes live) the "overflow" value:
https://github.com/mozilla/firefox-health-dashboard/blob/master/src/index.css#L255
I get this:
https://cl.ly/0L0B093c341j

If you select the <header> elemnt and check at the CSS rules for it you will notice that there are two background-colour properties applied to it (gray and green background).
If we add background-color *only* to the <h3> element we get this:
https://cl.ly/2L1Z1w311L2A

I think there might be two ways here:
* Make those CSS changes
* Modify the smiley png to have transparency outside of the face (leave the teeth white)

[...]

PR #62, opened 2018-03-07
Issue #43 - Support benchmarks with subbenchmarks

armenzg:
We are adding a new route for Quantum that allows showing a breakdown
for each subbenchmarks for a given benchmark.

You can test it by visiting https://http://health.graphics/quantum/linux64/ARES6

### DIFF
--- a/src/quantum/subbenchmarks.jsx
+++ b/src/quantum/subbenchmarks.jsx
@@ -1,0 +1,148 @@
+import { Component } from 'react';
+import { Route, Link } from 'react-router-dom';
+import propTypes from 'prop-types';
+import MetricsGraphics from 'react-metrics-graphics';
+import FontAwesomeIcon from '@fortawesome/react-fontawesome';
+import faLink from '@fortawesome/fontawesome-free-solid/faLink';
+import faExternalLinkAlt from '@fortawesome/fontawesome-free-solid/faExternalLinkAlt';
+
+import getData from '../utils/perfherder/subbenchmarks';
+
+const DEFAULT_PERCENTILE_THRESHOULD = 99;
+
+const Subbenchmarks = ({ match }) => (
+  <div>
+    <PerfherderContainer
+      suite={match.params.suite}
+      platform={match.params.platform}
+    />
+  </div>
+);
+
+Subbenchmarks.propTypes = {
+ match: propTypes.shape({}).isRequired,
+};
+
+const Graph = ({ data, name, url }) => (
+  <div key={name}>
+    <div className='black-bar'>
+      <a href={`#${name}`}>
+        <FontAwesomeIcon
+          id={name}
+          icon={faLink}
+          style={{ marginRight: '0.3em' }}
+        />
+      </a>
+      <a href={url} target="_blank">{name}</a>
+    </div>
+    <MetricsGraphics
+      key={name}
+      data={data}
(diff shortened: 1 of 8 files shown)


armenzg:
Hi @wlach do you have some feedback on this?

CC @stuartphilp @jmaher feedback wanted

The feature works like this:

- Visit https://platform-health-pr-62.herokuapp.com/quantum
- Go down to the "#3 Other benchmarks" section
- Click on the "Breakdown" link [1]
  - **TODO**: I have to adjust the CSS there to move the text to the right
- This will take you to the [breakdown view](https://platform-health-pr-62.herokuapp.com/quantum/linux64/ARES6) [2]

I've also added bookmarkibility:

- For instance https://platform-health-pr-62.herokuapp.com/quantum/linux64/ARES6#ML_firstIteration
  -  **TODO**: I've noticed the icon does not work; I will fix that

[1] 
image

[2]
image

jmaher:
the linking to a specific test doesn't work for me, but the rest does.

some general feedback:
1) I don't like the curvy lines- it makes it harder for me to see longer
term trends
2) we had a spike on march 6th- that adjusted all the Y axis scale- can we
find a way to set the scale on 95% of the data and mouse over for the
outliers?
3) I would like geomean to be at the top- that is the benchmark score

Otherwise- the flow from the platform health to the benchmark looks good.

src/quantum/subbenchmarks.jsx
```diff
@@ -0,0 +1,121 @@
...
+                    <a href={el.meta.url} target="_blank">{el.meta.test}</a>
+                  </div>
+                  <MetricsGraphics
+                    key={el.meta.test}
+                    data={[el.data]}
+                    {...{
+                      x_accessor: 'datetime',
+                      y_accessor: 'value',
+                      min_y_from_data: true,
+                      full_width: true,
```
wlach: I don't think the chart should be an area chart. Try setting `area: false` here.

I believe that should also take care of the curving issue that jmaher mentioned

wlach:
@armenzg check that, I think the weird curviness is due to interpolation. Try using a linear curvature instead:

https://github.com/mozilla/missioncontrol/blob/4c893c255c8f3dd623ff380c4c111b8980f8de94/frontend/ui/detailgraph.jsx#L25

armenzg:
Thanks for the suggestion!
I don't think it is much of an improvement.

I will try with just dots to see how it looks.

image

wlach:
That actually looks a lot better to me. But points or some kind of sliding confidence window may be even better. I guess you can experiment with what looks good to people.

armenzg:
I'm satisfied with what I have.

This is what I will be merging:
* Support 2nd title for Widget criteria
  * Cleaned up CSS issues for Widget criteria (See issue #45 )
* Support for subbenchmarks API ([/quantum/linux64/ARES6](https://platform-health-pr-73.herokuapp.com/quantum/linux64/ARES6))

I'm leaving some things out of this PR:
* #70 - Enable section with new benchmarks
  * Since we don't care to show Linux64 and it was only until we've Windows support
  * Enabled Linux64 benchmarks in the `enable_linux64_benchmarks` branch
  * You can see the deployment [here](https://platform-health-pr-73.herokuapp.com/quantum) 
* #71 - Confidence band

PR #73, opened 2018-03-19
DO NOT MERGE - Enable Linux64 benchmarks: ARES6 & others

armenzg:
This PR is simply to create a review app on Heroku.
We're waiting for Windows support on TaskCluster and this will serve as a testing ground until then.
We could have also used a staging app, however, it would be easy to forget and leave it around.
Once this PR is closed it will destroy the review app.

### DIFF
--- a/src/quantum/config.js
+++ b/src/quantum/config.js
@@ -1,0 +1,34 @@
+const CONFIG = {
+  regression: [
+    [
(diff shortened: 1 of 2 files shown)


sarah-clements:
This branch was updated in pr #86 and merged into master.

Issue #54, opened 2018-02-28
Distinguish Firefox versions on graphs


armenzg:
The Photon graphs show different colours for each Fx version [1].
The data coming from Telemetry has the data split like that already.

It would be nice if we could take the same approach for the Speedometer graphs [2] (it currently only shows an orange line).

We can handle this by moving data points into buckets according to the release dates.
We will need a data source to determine the dates or update them manually.

We could also add simple markers on the graph indicating the release dates.

This [versions API](https://health.graphics/api/release/versions) could have been the one, however, it seems to return 404. Filed as https://github.com/mozilla/firefox-health-backend/issues/15.

[1]
![image](https://user-images.githubusercontent.com/44410/36808691-d59f2990-1c93-11e8-9359-1dc0098f7911.png)
[2] 
![image](https://user-images.githubusercontent.com/44410/36808721-eba9238a-1c93-11e8-8bc0-542a9753f3f3.png)

armenzg:
We can probably easily be able to use [annotations](https://www.metricsgraphicsjs.org/examples.htm#annotations) from metrics-graphics.

PR #89, opened 2018-04-27
Bug #54 - distinguish firefox versions

sarah-clements:
add markers to metric-graphics speedometer graphs to denote firefox releases

endpoint changes are [here](https://github.com/mozilla/firefox-health-backend/pull/29)

### DIFF
--- a/src/components/awfy/speedometer.js
+++ b/src/components/awfy/speedometer.js
@@ -54,13 +54,29 @@ export default class Speedometer extends Component {
 
   state = {
     series: undefined,
+    markers: undefined,
   }
 
   componentDidMount() {
-    this.fetchPlotGraph(this.props);
+    this.fetchMarkerData();
   }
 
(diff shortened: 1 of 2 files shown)


src/components/awfy/speedometer.js
```diff
@@ -54,14 +54,30 @@ export default class Speedometer extends Component {
...
   }
 
   componentDidMount() {
-    this.fetchPlotGraph(this.props);
+    this.fetchMarkerData();
   }
 
   viewport: [0, 0];
 
+  transformMarkerData(data) {
```
wlach: Couldn't you just do something like:

```js
return data.map(d => ({ date: d.date, label: d.version }));
```

(maybe even dispensing with the helper function.

sarah-clements: Thanks for the suggestion

Issue #39, opened 2018-02-15
Add android release criteria dashboard


stuartphilp:
As discussed, and as a separate dashboard/link from the home page, we'd like a dashboard showing how things are tracking for android/geckoview throughout the year

PR #90, opened 2018-05-03
Bug #39 - Add android dashboard

sarah-clements:
create android route and graph and widget component; replace flow component with newer graph component


src/utils/helpers.js
```diff
@@ -0,0 +1,38 @@
...
+    const item1 = (desc ? b[sortBy] : a[sortBy]);
+    const item2 = (desc ? a[sortBy] : b[sortBy]);
+
+    if (item1 < item2) {
+      return -1;
+    }
+    if (item1 > item2) {
+      return 1;
+    }
+    return 0;
```
wlach: I don't think this helper method is really since it's only used in one place. Besides that, the logic could be consderably simplified:

* don't bother with the `desc` parameter (just call reverse if you want it in descending order)
* you can just subtract one element from the other instead of doing the comparison (i.e. `return item1 - item2`)

src/utils/helpers.js
```diff
@@ -0,0 +1,38 @@
+// separates an array of objects into the number of desired arrays of objects
+// (data structure provided by graphData) each containing a date and value;
+// converts date strings into date objects - as required for multi-line MG graphs
+export const transformGraphData = (graphData, data) => {
+  const finalData = [];
+  for (const key of Object.keys(graphData)) {
+    for (const item of data) {
+      graphData[key].push({ date: new Date(item.date), value: item[key] });
+    }
```
wlach: This method seems kind of strange to me. Is the graphData parameter somehow already populated with some data and we're adding more to it? It seems like we're calling this after getting some data from the server. I haven't delved into the details, but is there no way we couldn't just simplify this to a single statement (preferably inline if you're only going to use it once) which transforms the data as required? [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) is often handy if you want to build a dictionary type structure.

src/utils/helpers.js
sarah-clements: Good point - I'll change it to a one liner.

src/utils/helpers.js
sarah-clements: I'll push a refactored version with a clearer explanatory text.

sarah-clements:
@wlach let me know what you think of the changes - thanks!

Issue #94, opened 2018-05-07
Speedometer graph to link to Perfherder's autoland/m-i to discover alerts


armenzg:
I saw this baseline bump for Speedometer
https://cl.ly/1O1b421A1a1o
which lead me to this PH graph:
https://treeherder.mozilla.org/perf.html#/graphs?series=%5Bmozilla-central,797a2bfbbac2c632f4f354aff7677a3df3c749a3,1,1%5D&series=%5Bmozilla-central,be6032b74e5c65e733f105927222b0880c2e9822,1,1%5D&timerange=7776000

![image](https://user-images.githubusercontent.com/44410/39718125-59744044-5203-11e8-9696-1f42ada1b2b5.png)

Now, if we navigated to m-i/autoland instead of m-c we would actually get to see perf alerts that caused the regression (see the screenshot):
https://treeherder.mozilla.org/perf.html#/graphs?timerange=5184000&series=mozilla-central,1652331,0,1&series=mozilla-inbound,1643923,1,1&selected=mozilla-inbound,1643923,321856,440842331,1

![image](https://user-images.githubusercontent.com/44410/39718078-31f47700-5203-11e8-956a-c41c27761ee4.png)

I'm also going to reduce the timerange from 90 days to 30 days to help see through so many dots since I was unable to notice the dots with alerts at first glance:
https://bugzilla.mozilla.org/show_bug.cgi?id=1459623

PR #97, opened 2018-06-06
Issue #94 - Fix Quantum page crashing

armenzg:
Unfortunately, the quantum flow burn up API was returning an empty
array causing the GraphContainer to pass it down and that causing issues.

The API mentioned is this one:
https://firefox-health-backend.herokuapp.com/api/bz/burnup/?whiteboard=%5Bqf%3Ap1%5D%5Bqf%3Af61%5D

We should add error boundaries for widgets so the whole page does not get unmounted.

@sarah-clements I'm going to land this before you see it. Do you have any suggestions on other ways of fixing this? Or tests to be added? Or anything to be changed on the backend? Thanks!

### DIFF
--- a/src/components/graph-container.js
+++ b/src/components/graph-container.js
@@ -44,6 +44,10 @@ export default class GraphContainer extends React.Component {
   }
 
   parseData(data) {
+    if (data.length === 0) {
+      this.setState({ apiFailed: true });
+      return;
+    }
     const { checkStatus, keys, targetValue, targetLine } = this.props;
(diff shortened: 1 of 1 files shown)


armenzg:
Associated backend issue:
https://github.com/mozilla/firefox-health-backend/issues/31

sarah-clements:
I think the empty array that the endpoint is returning is the correct response if no data is found with the specified whiteboard tags, so I wouldn't set the state to apiFailed = true in parseData. But there should be conditions for checking the length of the array in parseData before transforming it. Sorry for missing that!

PR #100, opened 2018-06-06
Fix for quantum flow component

sarah-clements:
modify order of whiteboard tags and create noDataFound state for empty arrays

### DIFF
--- a/src/components/graph-container.js
+++ b/src/components/graph-container.js
@@ -18,6 +18,7 @@ export default class GraphContainer extends React.Component {
       data: null,
       status: null,
       apiFailed: false,
+      noDataFound: false,
     };
   }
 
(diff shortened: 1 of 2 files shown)


sarah-clements:
This is a fix pertaining to [this](https://github.com/mozilla/firefox-health-dashboard/pull/97) issue.
@armenzg 
The problem was two-fold:
1) the graphContainer component was not handling the case of an empty array (which would be the correct response if nothing was retrieved from bugzilla based on the query params) - I've updated the component state to handle this scenerio.

2) bugzilla (in the burnup endpoint) seems to be performing an exact match of any whiteboard tags entered as query params. So '[qf:f61][qf:p1]' sequence returns results, but '[qf:p1][qf:f61]' does not. Perhaps it'd be a good idea to revisit the use of bmo elasticsearch service. I had initially looked into this with the help of Kyle, but thought it wasn't necessary in the end (I initially thought either tag was necessary, not both, and thus needed 'or' logic). The ES query I had started on is [here](https://pastebin.com/h9vE7FWJ). If you use this, some of the endpoint logic will need to be changed because some of the bug metadata (creation_time, last_change_time, etc) is not named the same with ES. Also, Kyle mentioned some changes with regards to the ES service that was scheduled to happen in May (don't recall the specifics) that you'd want to check in about.

Issue #99, opened 2018-06-06
Add a first-blank-paint Photon perf graph


past:
In 61 we [added](https://bugzilla.mozilla.org/show_bug.cgi?id=1453782) a new perf measurement for the early blank window paint that has yet to ship to our release population. You can see it as the last graph in my dashboard:

https://astithas.com/perm/photon-perf/

It would be good to add this to the main Quantum dashboards as well.

armenzg:
I will look into this tomorrow

armenzg:
@past Could you please give me your feedback?
https://platform-health-pr-102.herokuapp.com/quantum/64

Clicking on the header of the graph will take you here:
https://telemetry.mozilla.org/new-pipeline/evo.html#!aggregates=median&arch=x86-64&cumulative=0&end_date=null&keys=&max_channel_version=nightly%252F62&measure=SIMPLE_MEASURES_BLANKWINDOWSHOWN&min_channel_version=nightly%252F59&processType=*&product=Firefox&sanitize=0&sort_keys=submissions&start_date=null&trim=0&use_submission_date=0

![image](https://user-images.githubusercontent.com/44410/41161941-5eae083e-6b02-11e8-8a31-5c6d3e8b9e9c.png)

past:
Looks great! The only odd thing is that it doesn't seem to be filtering for parent process only (same as First Paint), which is what these measurements are really about. Not a big deal though if it's too cumbersome to fix.

armenzg:
What would a fix look like? I don't quite understand.
What would a correct fetch to Telemetry look like?

past:
You could [pass](https://telemetry.mozilla.org/new-pipeline/evo.html#!max_channel_version=nightly%252F62&min_channel_version=nightly%252F60&sanitize=1&trim=1&use_submission_date=0&measure=SIMPLE_MEASURES_BLANKWINDOWSHOWN&product=Firefox&processType=parent) the `processType=parent` query argument or add it as a [filter](https://github.com/past/past.github.com/blob/master/perm/photon-perf/js/index.js#L70) if you are using TelemetryWrapper.

armenzg:
That filter is defined on the backend:
https://github.com/mozilla/firefox-health-backend/blob/photon/src/perf/telemetry_config.js#L41

I think there might be a bug on the frontend where the link is built incorrectly.
If I put your graph and the one produced in my PR I can see that they're plotting the same data (see screenshot).

[...]

PR #102, opened 2018-06-08
Issue #99 - Use new generic telemetry API + add blankWindowShown

armenzg:


armenzg:
The backend changes are in here:
https://github.com/mozilla/firefox-health-backend/pull/32
The new metric can be reached here:
https://firefox-health-backend-pr-32.herokuapp.com/api/perf/telemetry?architecture=x86-64&name=blankWindowShown

You can see the frontend deployment here:
https://platform-health-pr-102.herokuapp.com/quantum/64

This and the backend changes need to land at the same time.

src/telemetry/graph.js
```diff
@@ -7,17 +7,15 @@ import SETTINGS from '../settings';
...
-    let url = `${SETTINGS.backend}/api/perf/telemetry/${id}`;
-
-    if (Object.keys(queryParams).length > 0) {
-      url += `?${stringify(queryParams)}`;
-    }
+  async fetchPlotGraph(id, queryParams) {
+    let url = `${SETTINGS.backend}/api/perf/telemetry?`;
+    url += stringify({
+      name: id,
+      ...queryParams,
```
sarah-clements: Why would you need to pass in a copy of queryParams? Does stringify mutate it? I was trying to test this locally but hit CORS issues with the temp backend.

armenzg: You either need to `yarn start:local` or use the BACKEND env variable to point to https://firefox-health-backend-pr-32.herokuapp.com

I don't exactly know what you're referring.

Javascript's object destructuring is similar to doing this:
```
const fooObject = { name: id };
Object.entries(queryParams).forEach((key, value) => fooObject[key] = value);
```
Also, as part of the linting rules, I am not allowed to modify any parameters specified on a function's signature. In our case `id` and `queryParams` (referred signature: `fetchPlotGraph(id, queryParams)`).

sarah-clements: Sorry for the confusing statement - I thought I had edited my comment.  I did get the server running locally in the end. I initially had a strange caching issue with the BACKEND env variable change not being bundled and rendered despite it clearly showing up in the code.

As for the queryParams question, what I meant was, I was curious why you needed to use spread syntax  `...queryParams` - to copy properties of that object - rather than just passing the original object in to stringify. Does stringify mutate the original object?

Issue #106, opened 2018-06-14
Error boundaries per widget


armenzg:
To prevent a widget unhandled error to crash the page.

PR #107, opened 2018-06-15
Issue #106 - Wrap all components in generic error boundaries

armenzg:
This allows the UI to still be somewhat useful even if one component
produces an uncaught error.

You can test that this works by adding something that will throw an error in one of the components.
For instance insice of src/quantum/countdown.js I added the line `date.lenght();`.

I still have to give this one more pass to make sure *all* widgets are wrapped as well as the Android release criteria page.

@sarah-clements This is the PR about error boundaries I mentioned to you

### DIFF
--- a/src/components/criticalErrorMessage.jsx
+++ b/src/components/criticalErrorMessage.jsx
@@ -1,0 +1,12 @@
+const newIssue = 'https://github.com/mozilla/firefox-health-dashboard/issues/new';
+
+const CriticalErrorMessage = () => (
+  <p style={{ textAlign: 'center', fontSize: '1.5em' }}>
+    <span>There has been a critical error. We have reported it. If the issue is not fixed
+      within few hours please file an issue:</span>
+    <br />
+    <a href={newIssue} target="_blank" rel="noopener noreferrer">{newIssue}</a>
+  </p>
+);
+
+export default CriticalErrorMessage;
--- a/src/components/genericErrorBoundary.jsx
+++ b/src/components/genericErrorBoundary.jsx
@@ -1,0 +1,40 @@
+import Raven from 'raven-js';
+import { Component } from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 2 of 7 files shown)


src/components/criticalErrorMessage.jsx
```diff
@@ -0,0 +1,12 @@
+const newIssue = 'https://github.com/mozilla/firefox-health-dashboard/issues/new';
+
+const CriticalErrorMessage = () => (
+  <p style={{ textAlign: 'center', fontSize: '1.5em' }}>
+    <span>There has been a critical error. We have reported it. If the issue is not fixed
+      within few hours please file an issue:</span>
+    <br />
+    <a href={newIssue} target="_blank">{newIssue}</a>
```
eliperelman: You should get in the habit of using `rel="noopener noreferrer"` when using `a` tags to external links for security purposes (tab nabbing).

src/components/genericErrorBoundary.jsx
```diff
@@ -0,0 +1,40 @@
...
+        return <CriticalErrorMessage />;
+      }
+      return (
+        <div>
+          <p style={{ color: 'black' }}>Oops; something went wrong</p>
+        </div>
+      );
+    }
+    // eslint-disable-next-line react/prop-types
+    return (this.props.children);
```
eliperelman: No need to wrap in parens.

src/utils/componentEnhancers.jsx
```diff
@@ -0,0 +1,15 @@
+import GenericErrorBoundary from '../components/genericErrorBoundary';
+
+const wrapSectionComponentsWithErrorBoundaries = (sections = []) => {
+  const newSections = [];
```
eliperelman: This could be done with a map:

```jsx
return sections.map(section => ({
  ...section,
  rows: section.rows.map(row => row.map((component, index) => (
    <GenericErrorBoundary key={index}>{component}</GenericErrorBoundary>
  ))
}))
```

PR #118, opened 2018-07-04
Add Android V2 route which uses Nimbledroid data

armenzg:
It is configured to use this WIP backend:
https://firefox-health-backend-pr-36.herokuapp.com

### DIFF
--- a/src/components/NimbledroidWidget/index.jsx
+++ b/src/components/NimbledroidWidget/index.jsx
@@ -16,12 +16,15 @@ export default class NimbledroidWidget extends Component {
     const labels = Object.keys(profile.data);
     const data = labels.map(product => profile.data[product]);
     const target = this.props.targetRatio * profile.lastDataPoints.focus;
+    const status = profile.lastDataPoints.klar / profile.lastDataPoints.focus;
 
     return (
       <Widget
         // This classname is to undo what .criteria-widget:not:first-child sets
         className={'no-left-margin'}
         title={profile.title}
+        targetStatus={status < 1 ? 'pass' : 'fail'}
+        status={status < 1 ? 'green' : 'yellow'}
       >
         <div>
           <div className='legend' ref={ele => this.legendTarget = ele} />
--- a/src/components/NimbledroidWidget/index.jsx
+++ b/src/components/NimbledroidWidget/index.jsx
@@ -1,0 +1,50 @@
+import { curveLinear } from 'd3';
+import PropTypes from 'prop-types';
+import { Component } from 'react';
+import MetricsGraphics from 'react-metrics-graphics';
+
+import Widget from '../../quantum/widget';
+
+export default class NimbledroidWidget extends Component {
+  constructor(props) {
+    super(props);
+    this.legendTarget = null;
+  }
+
+  render() {
+    const { profile } = this.props;
+    const labels = Object.keys(profile.data);
+    const data = labels.map(product => profile.data[product]);
+
+    return (
+      <Widget
+        // This classname is to undo what .criteria-widget:not:first-child sets
+        className={'no-left-margin'}
+        title={profile.title}
+      >
+        <div>
+          <div className='legend' ref={ele => this.legendTarget = ele} />
+          {profile &&
+            <MetricsGraphics
+              width={600}
+              height={300}
+              data={data}
+              x_accessor='date'
+              y_accessor='value'
+              legend={labels}
+              legend_target={this.legendTarget}
+              interpolate={curveLinear}
+            />
+          }
+        </div>
+      </Widget>
+    );
+  }
+}
+
+NimbledroidWidget.propTypes = {
+  profile: PropTypes.shape({
+    title: PropTypes.string.isRequired,
+    data: PropTypes.shape({}).isRequired,
+  }),
+};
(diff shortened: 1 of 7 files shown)


armenzg:
@wlach Hi Will, if you have some time I'm having some performance issues with metrics-graphics.
Here's a screenshot of it:
https://www.dropbox.com/s/9eegq7nxg2rd1f6/Screenshot%202018-07-09%2016.05.04.png?dl=0

I wonder what is the right way of doing this. Since I have over a 100 charts to plot.

wlach:
@armenzg how many datapoints are you plotting per chart? that's the main driver of how slow/fast metricsgraphics is gonna be

src/utils/BackendClient/NimbledroidApiHandler.js
```diff
@@ -0,0 +1,103 @@
...
+    status !== 'fail' &&
+    scenario_name.startsWith('customFlow') &&
+    scenario_name.includes('http')
+  ) {
+    return true;
+  }
+  return false;
+};
+
+const transformedDataForMetrisGraphics = (nimbledroidData) => {
```
wlach: Similarly, I would just put this function inline in the one place where it is used, maybe with a quick comment like `// tranform data into format metrics graphics expects`

src/utils/BackendClient/NimbledroidApiHandler.js
```diff
@@ -0,0 +1,103 @@
...
+const scenarioShouldBeConsidered = ({ scenario_name, status }) => {
+  if (
+    status !== 'fail' &&
+    scenario_name.startsWith('customFlow') &&
+    scenario_name.includes('http')
+  ) {
+    return true;
+  }
+  return false;
+};
```
wlach: I don't think defining this as an external function helps with clarity. I would put it inline, maybe with a reference to any documentation on why this criteria we have available (or just say why in the comment, if we have no such documentation).

src/views/AndroidV2/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+        className={cx('summary')}
+      >
+        <div className='row'>
+          {Object.keys(nimbledroidData).length > 0 &&
+            <NimbledroidGraphs
+              nimbledroidData={nimbledroidData}
+            />
+          }
+          {Object.keys(nimbledroidData).length === 0 &&
+            <h2>The backend is loading data from Nimbledroid. Thanks for waiting.</h2>
```
wlach: Adding some kind of spinner widget here might be nice, if you expect this operation to take more than a second.

src/components/NimbledroidWidget/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+        targetStatus={status < 1 ? 'pass' : 'fail'}
+        status={status < 1 ? 'green' : 'yellow'}
+      >
+        <div
+          style={{
+            display: 'grid',
+            gridTemplateRows: '0.1em 1fr',
+          }}
+        >
+          <div className='legend' ref={ele => this.legendTarget = ele} />
```
wlach: I suspect this logic is failing, and that's why you're not seeing the legend until you reload the page.

src/views/AndroidV2/index.jsx
armenzg: Linear query works for you?
https://material-ui.com/demos/progress/#linear-query

src/utils/BackendClient/NimbledroidApiHandler.js
armenzg: IMHO I beg to disagree.
The current `fetchData` function is clear that it only does two things: it fetches and transforms the data. The reader does not need to look at 40 extra lines of code to determine if anything else is happening besides a transformation of the fetched data. The name of the function is self-explanatory.

I'm trying to follow some clean code pattern I've studying.
If you have no strong objections I will leave it as such.

wlach: These questions don't have clear answers I suppose. I know some people prefer to have lots of little functions (even if they're only called once) as a means to decompose the structure of their program into some kind of logical hierarchy. Personally I don't find that helpful, I'd rather just see the entire control flow in one place for something like this since I can just read it from top to bottom and be done with it (you can find a longer argument on this here: https://medium.com/@copyconstruct/small-functions-considered-harmful-91035d316c29) . But I won't insist here if you feel strongly you want to code in this style.

src/components/NimbledroidWidget/index.jsx
armenzg: I've used [this logic](https://github.com/mozilla/firefox-health-dashboard/blob/acdc71e815ca50b93707d401cfd019b44471b194/src/components/graph-container.js#L86-L87) succesfully in the past. You can see it working in [Android page](https://health.graphics/android).

It's not on reload when the legends will show up. It's on window resize.

If I use `(profile && this.legendTarget)` [1] as in the code mentioned above I won't get the graphs until I resize the window. I left `&& this.legendTarget` so I could at least guarantee the drawing of the graph even if the labels are not drawn.
I need to get a smaller test case.

[1]
```javascript
<div className='legend' ref={ele => this.legendTarget = ele} />
{(profile && this.legendTarget) &&
```

wlach: Sorry, yeah, I meant resize not reload. I'm not sure, but it could be there is a race condition with this type of logic. Perhaps in the other example, the bug is present but not observable because things happened in a different order.

src/views/AndroidV2/index.jsx
wlach: Sure, whatever looks good to you!

PR #127, opened 2018-07-18
Better errors for Nimbledroid page

armenzg:


src/utils/BackendClient/NimbledroidApiHandler.js
```diff
@@ -110,6 +110,12 @@ class NimbledroidApiHandler {
     const productsData = await Promise.all(
       this.products.map(async product => this.fetchProductData(product)),
     );
+    if (
+      Object.keys(productsData[0]).length === 0 ||
+      Object.keys(productsData[1]).length === 0
+    ) {
+      throw Error('The backend has no data available.');
```
sarah-clements: I'm curious, why are you throwing an error message instead of returning an http status code to be handled by the frontend? If you need to do it this way, it'd be a good idea to have a comment explaining that any change in the message's wording needs to also be changed in the frontend or it'll break the logic.

src/views/AndroidV2/index.jsx
```diff
@@ -20,12 +20,29 @@ export default class AndroidV2 extends Component {
       const nimbledroidData = await this.client.getData('nimbledroid');
       this.setState({ nimbledroidData });
     } catch (e) {
-      this.setState({
-        errorMessage: 'Failed to fetch data from the backend. We have reported it.',
-      });
-      if (e.message === 'Failed to fetch' && process.env.NODE_ENV === 'production') {
-        Raven.captureMessage('Failed to fetch the Nimbledroid data from the backend.');
-        Raven.captureException(e);
+      if (e.message === 'Failed to fetch') {
```
sarah-clements: Same thing as the point I mentioned above.

armenzg:
You've got a good point. I've fixed this on the backend.

PR #128, opened 2018-07-19
Remove App component controlling resizing

armenzg:
I've tried using the app without it and I can't notice any issues.


src/routes.js
```diff
@@ -13,59 +13,21 @@ import AndroidV2 from './views/AndroidV2';
...
-            <Route path='/' exact component={Home} />
-            <Route path='/crashes' exact component={ReleaseCrashes} />
-            <Route path='/crashes/beta' component={BetaCrashes} />
-            <Route path='/status' component={Status} />
-            <QuantumRoutes />
-            <Route component={NoMatch} />
-          </Switch>
-        </App>
+        <Switch>
+          <Route path='/android' exact component={Android} />
```
eliperelman: The idiomatic way to quote props is double quotes.

src/routes.js
```diff
@@ -13,59 +13,21 @@ import AndroidV2 from './views/AndroidV2';
...
+          <Route path='/android/v2' exact component={AndroidV2} />
+          <Route path='/' exact component={Home} />
```
eliperelman: You can avoid the usage of the `exact` prop if you rearrange these from most-specific to least specific:

```jsx
<Route path="/android/v2" component={AndroidV2} />
<Route path="/android" component={Android} />
<Route path="/crashes/beta" component={BetaCrashes} />
<Route path="/crashes" component={ReleaseCrashes} />
<Route path="/status" component={Status} />
<Route path="/devtools" component={Devtools} />
<QuantumRoutes />
<Route path="/" component={Home} />
<Route component={NoMatch} />
```

src/routes.js
```diff
@@ -13,59 +13,21 @@ import AndroidV2 from './views/AndroidV2';
...
+          <Route path='/crashes' exact component={ReleaseCrashes} />
+          <Route path='/crashes/beta' component={BetaCrashes} />
+          <Route path='/status' component={Status} />
+          <Route path='/devtools' component={Devtools} />
+          <QuantumRoutes />
```
eliperelman: Why do these routes exist in a different file?

armenzg: It pre-dates me.

armenzg:
I've addressed all you pointed out. Thanks!

armenzg:
Thank you so much!

PR #141, opened 2018-09-13
Use the new Nimbledroid API

armenzg:
Version 2 of the API has Json compression and reduced data.


armenzg:
There's no compression happening, thus, downloading over 6MB of data:
image

Fetching from new API only downloads ~1/12 of the amount of data:
image

Issue #155, opened 2018-09-27
Add GeckoView query URLs with badge


armenzg:
I have the following information to add.

* P1 GeckoView Bugs:

https://bugzilla.mozilla.org/buglist.cgi?v4=P1&j_top=OR&f1=status_whiteboard&o3=equals&v3=GeckoView&o1=anywordssubstr&resolution=---&f4=priority&query_format=advanced&f3=component&o4=equals&f2=OP&v1=%5Bgeckoview%3Ap1%5D%2C%5Bgeckoview%3Afocus%3Ap1%5D%2C%5Bgeckoview%3Aklar%3Ap1%5D%2C%5Bgeckoview%3Afxr%3Ap1%5D%2C%5Bgeckoview%3Acrow%3Ap1%5D

* GeckoView Backlog Bugs:

https://bugzilla.mozilla.org/buglist.cgi?v4=P2%2CP3&j_top=OR&f1=status_whiteboard&o3=equals&v3=GeckoView&o1=anywordssubstr&resolution=---&f4=priority&query_format=advanced&f3=component&o4=anywordssubstr&f2=OP&v1=%5Bgeckoview%3Ap2%5D%2C%5Bgeckoview%3Afocus%3Ap2%5D%2C%5Bgeckoview%3Aklar%3Ap2%5D%2C%5Bgeckoview%3Afxr%3Ap2%5D%2C%5Bgeckoview%3Acrow%3Ap2%5D%2C%5Bgeckoview%3Ap3%5D%2C%5Bgeckoview%3Afocus%3Ap3%5D%2C%5Bgeckoview%3Aklar%3Ap3%5D%2C%5Bgeckoview%3Afxr%3Ap3%5D%2C%5Bgeckoview%3Acrow%3Ap3%5D

* Untriaged GeckoView Bugs:

https://bugzilla.mozilla.org/buglist.cgi?keywords=meta&keywords_type=nowords&o1=nowordssubstr&o4=anywordssubstr&v1=%5Bgeckoview%5D%2C%5Bgeckoview%3Afocus%5D%2C%5Bgeckoview%3Aklar%5D%2C%5Bgeckoview%3Afxr%5D%2C%5Bgeckoview%3Acrow%5D&v4=--%2CP4&f1=status_whiteboard&o3=equals&v3=GeckoView&resolution=---&f4=priority&query_format=advanced&f3=component&f2=OP

PR #168, opened 2018-10-03
Issue #155 - Add Bugzilla query container with bug count badges to the Android health page

armenzg:


src/utils/bugzilla/generateBugzillaUrls.jsx
```diff
@@ -0,0 +1,46 @@
+import { stringify } from 'query-string';
+import fetchJson from '../fetchJson';
+
+const BZ_HOST = 'https://bugzilla.mozilla.org/buglist.cgi';
+const BZ_REST_API = 'https://bugzilla.mozilla.org/rest/bug';
+
```
sarah-clements: Might be more useful to put these in a `constants.js` file and import them.

armenzg: Because I believe they're module-level values rather than app-level I will keep them here.

Thanks for the review!

PR #171, opened 2018-10-09
Test Nimbledroid containers

armenzg:
Perhaps there's a way to use a plugin to not take a snapshot of a component until `componentDidMount` completes, however, this tricks the components to have the data from the get-go.

I also fix the drilldown situation where the whole Android health page would be displayed rather than just a graph.

### DIFF
--- a/src/containers/NimbledroidSummaryTable/index.jsx
+++ b/src/containers/NimbledroidSummaryTable/index.jsx
@@ -23,22 +23,22 @@ class NimbledroidSummaryTable extends Component {
     showSites: false,
   };
 
-  async componentDidMount() {
-    const { configuration, nimbledroidData, handleError } = this.props;
-    const { products } = configuration;
-
-    try {
-      const data = nimbledroidData || await fetchNimbledroidData(products);
-      this.setTableContents(data, configuration);
-    } catch (error) {
-      handleError(error);
+  constructor(props) {
+    super(props);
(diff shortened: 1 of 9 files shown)


src/containers/NimbledroidSummaryTable/index.jsx
```diff
@@ -23,22 +23,32 @@ class NimbledroidSummaryTable extends Component {
     showSites: false,
   };
 
+  constructor(props) {
+    super(props);
+    const { nimbledroidData } = this.props;
+    if (nimbledroidData) {
+      this.state = { ...(this.generateData(this.props)) };
+    }
```
sarah-clements: So this is basically receiving empty data along with other props? Seems a bit odd, but if you need to do it this way it'd be good to add a comment explaining why. :)

armenzg: Pretty much!

It is not for testing purposes. I wonder if I actually had a good reason or pattern or if I'm being plain lazy.

I will change it. Thanks for your review!

armenzg:
Hi @sarah-clements 
I noticed an issue we didn't catch.
When I changed the code, I pulled the `await` outside of the try/catch block.
I noticed a warning on the tests saying that a promise was not being handled within a try/catch block.
I just wanted to mention it in case you wanted to see what I mean.

PR #185, opened 2018-10-16
Compare Klar/GV against Chrome beta instead Focus/WebView

armenzg:


armenzg:
Unfortunately, comparing against klar/GV has quite the regression:
image

PR #193, opened 2018-10-23
WIP Bugzilla burndown

armenzg:


src/utils/bugzilla/advancedSearchToRestApi.js
```diff
@@ -0,0 +1,13 @@
+const TRANSFORM_FIELD = {
+    chfieldfrom: 'creation_time',
+};
+
+const advancedSearchToRestApi = parameters => (
+    Object.keys(parameters).reduce((result, key) => {
+        const newKey = TRANSFORM_FIELD[key] ? TRANSFORM_FIELD[key] : key;
```
sarah-clements: I think you could achieve the same result by doing this instead `TRANSFORM_FIELD[key] || key'

src/utils/bugzilla/chartJs/burndownDataFormatter.js
```diff
@@ -0,0 +1,52 @@
...
+            ...generateLineChartStyles(SETTINGS.colors[index]),
+            data: bugCountPerDay,
+            label,
+        });
+    });
+
+  return newData;
+};
+
+export default bugzillaFormatter;
```
sarah-clements: I think it'd be easier to trace this in files if you keep the name consistent with the filename and how it's named when imported.

src/utils/bugzilla/chartJs/burndownDataFormatter.js
```diff
@@ -0,0 +1,52 @@
+import generateLineChartStyles from '../../chartJs/generateLineChartStyles';
+import SETTINGS from '../../../settings';
+
+const bugsByCreationDate = (bugs, startDate) => {
+    // Count bugs created on each day
+    const byCreationDate = bugs.reduce((result, { creation_time }) => {
+        const date = creation_time.substring(0, 10);
+        // startDate enables counting into a starting date all previous data points
+        const actualDate = (startDate && date < startDate) ? startDate : date;
```
sarah-clements: Could you clarify how this logic works: `startDate && date < startDate`?

src/utils/bugzilla/chartJs/burndownDataFormatter.js
armenzg: That was a mistake. Thanks for catching it!

src/utils/bugzilla/advancedSearchToRestApi.js
armenzg: One day my brain will actually absorb this syntax. Thanks!

src/utils/bugzilla/chartJs/burndownDataFormatter.js
armenzg: I can understand why the parenthesis makes it easier for our brains to understand what is going on. I will add it.

PR #197, opened 2018-10-26
Use modern Perfherder component in the Quantum pages

armenzg:


armenzg:
Hi @sarah-clements,
This change modifies the Quantum pages to use PerfherderGraphContainer instead of the Perfherder component. It is a switch from a component that was written before my days using SVG and fetching performance data from the Health Backend to using ChartJs and fetching the data from Perfherder directly.

Later today I will release an official `perf-goggles` release instead of the beta we're using on this PR.

I'm intentionally dropping the feature of turning a widget green/yellow. Stuart and I agreed that it made no sense anymore.

We're loosing the hyperlinks to Perfherder which I would like to add back soon. I want to land this regardless if I have it ready or not.

Deployment:
https://deploy-preview-197--firefox-health-dashboard.netlify.com/quantum/64

Before:
image

After:
image

armenzg:
@sarah-clements I have released version 3.0.0 of perf-goggles which I'm now using in this PR.

armenzg:
Thanks for pointing that out! I will follow up.

PR #205, opened 2018-11-19
WIP: attempt to use ChartJS

klahnakoski:


klahnakoski:
this does not work well

image

1. The data is ther, you can see me hovering over a datapoint.  Where can I get reasonable default styling parameters for `ChartJsWrapper`?  I thought `generateOptions()` would work.
2. Is there a good link for ChartJS config options?  Maybe something like a JSON schema?  I use https://www.chartjs.org/docs/latest/ but I must do a lot of clicking to get the details
3. Can we add more to the chart title, like the old version has a link to Perfherder?

src/quantum/subbenchmarks.jsx
```diff
@@ -38,6 +41,26 @@ const Graph = ({ data, name, url }) => (
         />
       </a>
     </div>
+    <ChartJsWrapper
+      title={name}
```
armenzg: Don't pass a title.
You already have one with PerfherderGraphContainer:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/containers/PerfherderGraphContainer/index.jsx#L35

src/quantum/subbenchmarks.jsx
```diff
@@ -38,6 +41,26 @@ const Graph = ({ data, name, url }) => (
...
+      type='line'
+      height={80}
+      options={generateOptions()}
```
armenzg: Use `getPerferherderData` and you can pass the `options` it generates:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/containers/PerfherderGraphContainer/index.jsx#L30

src/quantum/subbenchmarks.jsx
```diff
@@ -38,6 +41,26 @@ const Graph = ({ data, name, url }) => (
...
+      data={{
```
armenzg: Use `getPerfherderData` and you can pass the `data` it generates.

src/quantum/subbenchmarks.jsx
```diff
@@ -38,6 +41,26 @@ const Graph = ({ data, name, url }) => (
...
+                y: value,
+              };
+            }),
+            label: name,
+          };
+        }),
+      }}
+    />
+
     <MetricsGraphics
```
armenzg: Remove `MetricsGraphics`

src/quantum/subbenchmarks.jsx
klahnakoski: from what I know `getPerferherderData` can not get subtest data.  Is there some other place that the styting information can come from?

src/quantum/subbenchmarks.jsx
klahnakoski: yes, I will remove metrics graphics when the other is working; I just need something to show to prove it I did not break other stuff

src/quantum/subbenchmarks.jsx
klahnakoski: looking at getPerfherderData: it seems to have changed its argument structure a bit since last time.  But it still requires I know the frameworkID, option, and project.  The subbenchmark code only knows suite and platform.

klahnakoski:
Attempting something simpler

PR #207, opened 2018-11-23
Add Perfprobes to Quantum pages

stuartphilp:
@armenzg front end part to add more perf probes, also moved things around a bit just to organize the 2 quantum pages

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -70,7 +70,7 @@
  }
  .dashboard-title .dashboard-note,.dashboard-title h1 small {
   padding-left:.75rem;
-  color:#d1d2d3
+  color:#000
  }
  .dashboard-title .dashboard-source {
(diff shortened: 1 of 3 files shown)


klahnakoski:
`src/quantum/index-32bit.js` and `src/quantum/index-64bit.js` are nearly identical except for 

* "windows7-32" replaces "windows10-64"
* "quantum64QueryParams" replaces "quantum32QueryParams"
* a few other places where "32" replaces "64"

Given that there is code at the bottom of these files, I would assume these two files should be one.

armenzg:
@klahnakoski your observations are correct. In the future, it will ideal to reconcile these two files and use parameters to change between the two architectures. Once we switch to the newer React components it will be easier to reunite them. For now it is easier to just add the same info on both pages.

Issue #204, opened 2018-11-19
Update dashboard for new tp6-3, tp6-4, tp6-5, and tp6-6 suites


rwood-moz:
With Bug 1503990 we will be adding new Raptor tp6-3, tp6-4, tp6-5, and tp6-6 suites, each of which will be running on both Firefox and Chromium. Details:

raptor-tp6-3:
- [raptor-tp6-imdb-firefox]
- [raptor-tp6-imgur-firefox]
- [raptor-tp6-wikia-firefox]
- [raptor-tp6-imdb-chrome]
- [raptor-tp6-imgur-chrome]
- [raptor-tp6-wikia-chrome]

raptor-tp6-4:
- [raptor-tp6-bing-firefox]
- [raptor-tp6-yandex-firefox]
- [raptor-tp6-bing-chrome]
- [raptor-tp6-yandex-chrome]

raptor-tp6-5:
- [raptor-tp6-apple-firefox]
- [raptor-tp6-microsoft-firefox]
- [raptor-tp6-apple-chrome]
- [raptor-tp6-microsoft-chrome]

raptor-tp6-6:
- [raptor-tp6-reddit-firefox]
- [raptor-tp6-reddit-chrome]

If they could be added to the dashboard that would be appreciated, thank you kindly!

klahnakoski:
I assume these should be added to "src/quantum/index-32bit.js" and its 64bit sibling.

Is there a logic behind the pairing I see in the chart layout?  For example **Tp6: Facebook** is beside **Tp6: Google**.  Or is being done instead of a flow layout? https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/quantum/index-32bit.js#L136..L187

armenzg:
There's no specific logic behind the sorting.

armenzg:
@klahnakoski let me know if you need another issue

PR #208, opened 2018-11-26
WIP: More charts

klahnakoski:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/204

### DIFF
--- a/src/quantum/index-32bit.js
+++ b/src/quantum/index-32bit.js
@@ -3,6 +3,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import { parse } from 'query-string';
+import { _ } from 'lodash/fp';
 import Dashboard from '../dashboard';
 import Perfherder from './perfherder';
 import Countdown from './countdown';
@@ -10,9 +11,10 @@ import TelemetryContainer from '../telemetry/graph';
 import SETTINGS from '../settings';
 import { quantum32QueryParams, flowGraphProps, statusLabels } from './constants';
 import GraphContainer from '../components/graph-container';
-import CONFIG from './config';
+import CONFIG, { PAGES } from './config';
 import wrapSectionComponentsWithErrorBoundaries from '../utils/componentEnhancers';
 import PerfherderGraphContainer from '../containers/PerfherderGraphContainer';
+
 
 export default class QuantumIndex32 extends React.Component {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


src/quantum/index-32bit.js
```diff
@@ -3,17 +3,19 @@ import React from 'react';
...
+import { _ } from 'lodash/fp';
 import Dashboard from '../dashboard';
 import Perfherder from './perfherder';
 import Countdown from './countdown';
 import TelemetryContainer from '../telemetry/graph';
 import SETTINGS from '../settings';
 import { quantum32QueryParams, flowGraphProps, statusLabels } from './constants';
 import GraphContainer from '../components/graph-container';
-import CONFIG from './config';
+import CONFIG, { PAGES } from './config';
```
klahnakoski: @armenzg I am not sure how to import

src/quantum/index-32bit.js
```diff
@@ -3,17 +3,19 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import { parse } from 'query-string';
+import { _ } from 'lodash';
```
klahnakoski: this appears to work in the debugger....

src/quantum/index-32bit.js
```diff
@@ -108,326 +110,38 @@ export default class QuantumIndex32 extends React.Component {
...
-                    suite: 'tsvgx',
-                  },
-                },
-              ]}
-            />,
-          ],
-        ],
+        title: '#3 Pages',
+        rows:
+          _
```
klahnakoski: ... but `_` is not recognized:

```
index-32bit.js?4211:114 Uncaught TypeError: Cannot read property 'chain' of undefined
    at ProxyComponent.render (index-32bit.js?4211:114)
```

klahnakoski: `import { _ as lodash } from 'lodash';`   does not help

klahnakoski:
@armenzg  The current problem is this odd warning:

```
react-dom.development.js?61bb:512 Warning: Each child in an array or iterator should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.
    in svg (at perfherder.js:183)
    in div (at widget.js:83)
    in div (at widget.js:73)
    in Widget (at perfherder.js:204)
    in PerfherderWidget (at index-32bit.js:99)
    in ErrorBoundary (at componentEnhancers.jsx:8)
    in div (at index-32bit.js:166)
    in div (at dashboard.js:48)
    in div (at dashboard.js:38)
    in Dashboard (at index-32bit.js:206)
    in QuantumIndex32 (created by Route)
    in Route (at routes.js:22)
    in Switch (created by Routes)
    in Router (created by BrowserRouter)
    in BrowserRouter (at routes.js:16)
    in Routes (at src/index.js:20)
    in ErrorBoundary
    in AppContainer (at src/index.js:18)
```

klahnakoski:
@armenzg   Ready for review

klahnakoski:
nope, not ready

klahnakoski:
@armenzg NOW it is ready

src/quantum/index-32bit.js
```diff
@@ -108,326 +109,39 @@ export default class QuantumIndex32 extends React.Component {
...
-                    option: 'pgo',
-                    project: 'mozilla-central',
-                    suite: 'tsvgx',
-                  },
-                },
-              ]}
-            />,
-          ],
-        ],
+        title: '#3 Pages',
```
armenzg: Why the notion of pages?
It's only the tp6 jobs that load pages. All the other benchmarks do other things than loading pages.
Please use 'benchmarks'.

src/quantum/index-32bit.js
```diff
@@ -108,326 +109,39 @@ export default class QuantumIndex32 extends React.Component {
...
+                _
+                .chain(series)
+                  .map((s) => {
+                    return {
+                      label: s.label,
+                      seriesConfig: s,
+                    };
+                  })
+                  .value()
+              }
```
armenzg: If you do this in the config:
```javascript
const CONFIG = {
   'Tp6: YouTube': {
      title='Tp6: YouTube',
      series=[
                {
                  label: 'Firefox',
                  seriesConfig: {
                    frameworkId: 10,
                    platform: 'windows7-32',
                    option: 'pgo',
                    project: 'mozilla-central',
                    suite: 'raptor-tp6-youtube-firefox',
                  },
                },
                {
                  label: 'Chrome',
                  seriesConfig: {
                    frameworkId: 10,
                    platform: 'windows7-32-nightly',
                    option: 'opt',
                    project: 'mozilla-central',
                    suite: 'raptor-tp6-youtube-chrome',
                  },
        ],
}
```
you can do the following here:
```javascript
Object.entries(config).map((key, value) => (
  <PerfherderGraphContainer
     key={key}
     ...value
   />
)
```

If you do not want to have 2 configs you can then remove `platform` from the configs and insert it at this point like this:
```javascript
Object.entries(config).map((key, value) => {
  const series = Object.assign(value, {});
  // TODO: Iterate through series and add the platform property
  // if the suite contains 'chrome' then also append '-nightly'
  return (<PerfherderGraphContainer
     key={key}
     title={value.title}
     series={series}
   />);
}
```

There's no need to have two graphs per row. We can just set a fix width and rely on flexing for this.
You can use this guide for help:
https://css-tricks.com/snippets/css/a-guide-to-flexbox/

src/quantum/index-32bit.js
klahnakoski: OK

src/quantum/index-32bit.js
klahnakoski: Putting the structured call pattern required for `<PerfherderGraphContainer>` into `CONFIG` only makes things more complicated: The information is not anonymous, it is not inlined, it has a name; which is another layer of indirection.

The `CONFIG` is an attempt to separate data from layout and style.  It has a few benefits over the  deep structured form used by `<PerfherderGraphContainer>`:

1. This form is centralized, and reduces the risk that the 32bit and 64bit pages deviate in some way.  It is step towards merging those two pages.
2. The `CONFIG` data is flatter, so it fits in a table, and it can be formatted to reveal its redundancy, *or important exceptions*. The deep structured form does not reveal these patterns or exceptions unless there is meticulous accounting of all the charts; even then something may be missed: did one of the charts used a special config?  This accounting is difficult using the `<PerfherderGraphContainer>` format because it spans hundreds of lines, and is mostly boilerplate with little signal. 
3. With `CONFIG` being flat, the patterns and redundancy is easy to spot: `CONFIG` shows off the fact that all the Chrome tests are run on one config, and Firefox is all run on another, except for the test in `frameworkId==1`, which appear to run on `['e10s', 'stylo']`.  We can take further steps to remove functional data dependencies to make it simpler.  This redundancy is not removed in this pull request, but it is a start to reveal there are patterns in the data. 
4. The `CONFIG` object is data, and the choice of data (the choice of tests shown) may have a pattern:  If we know this pattern then we can write some code that pulls all the tests based on a rule, rather than manually maintaining this `CONFIG`.  When we pull this `CONFIG` data dynamically, then the code need only provide simple, flat, data independent of the `<PerfherderGraphContainer>` calling conventions. 
5. With the data separate from the layout, we can see `<PerfherderGraphContainer>` is just one link in a chain of conversion functions:
    * `perf-goggles.queryPerformanceData()` transforms data, which gets passed to
    * `getPerfherderData()` which transforms the data and gets passed to
    * `perfherderFormatter()` which transforms the data and passes it to  
    * `<PerfherderGraphContainer>` which annotates the data with 
    * `generateOptions()` and then gets pass to 
    * `<ChartJsWrapper>` which transforms the data, and passes it to 
    * `<Chart>` which, ostensibly, does the drawing

    This chain, was hard to follow.  Each named function alludes to the fact some other code, not found in this chain, may need its transformation. But that is not the case: this is an exclusive chain.  At the very least this chain should be made from anonymous functions, or limited-scoped named, functions, or merged into just one function, so the API surface area is minimized.  
6. By separating `CONFIG` as data, we can start pointing to the major features of any  BI/dashboarding tool; which helps with code navigation:  Business rules, showing what tests are important, and what platforms those test run on, are in CONFIG.  The layout and style is  demonstrated by the `index32bit.js`, and pages in general.  A Chart API uses a standard data format that is consumed by `<ChartJsWrapper>`, or any other charting lib.  Storage accessed by  `queryPerformanceData()` (a Query API) which should be returning something close to perfherder's native format.
image

The `src/quantum/config.js` looked like a good spot to start adding the quantum business rules.

klahnakoski:
> please remove the disabling of the linting

This depends on our conversation about `CONFIG` in general. **If** `CONFIG` is allowed, then do you have a suggestion how to store the data to maintain the columnar form?  Columns are important for visually identifying data discrepancies. 

> and remove the usage of lodash.

Hmm, the code will get longer, and harder-to-read with clunky map/reduce. Lodash is part of the family of list-comprehensions; which are more expressive than map/reduce, but less expressive than a generalized query language.  JQuery and D3 are two more examples in Javascript, Python has list comprehensions. C# has linq. These exist because they simplify data transformation. This codebase does a lot of data transformation; it could benefit from using chained list comprehensions that lodash provides.

List comprehensions do require a learning curve to use, but they read easier than loops and maps. Demanding more list comprehensions will force better form on the data the is being manipulated: Right now there is no pattern to the data: Every step in the transformation pipeline is a free-form transformation from one type of deep nested inner objects to another: The code looks complicated only because the data structures are complicated. Moving to list comprehensions demands some flattening of data, which makes it easier to reason about, makes transformations simpler, and results in easy-to-read list comprehension chains.

> Also rebase the branch.

I will do this after more changes are made:  I am sad that this repo is not using [gitflow](https://datasift.github.io/gitflow/IntroducingGitFlow.html)

klahnakoski:
> There's no need to have two graphs per row. We can just set a fix width and rely on flexing for this.

Ok, I did not know why the dashboards were layout `mod(2)` for a specific reason; maybe moving to flow layout proved to be "too flexible" and unpredictable.

armenzg:
> > please remove the disabling of the linting
> 
> This depends on our conversation about `CONFIG` in general. **If** `CONFIG` is allowed, then do you have a suggestion how to store the data to maintain the columnar form? Columns are important for visually identifying data discrepancies.
> 
The value of consistency accross the board is far more important than visual inspection.
Even columnar form has issues of missing things.
You could even follow this form (and only disable the linting rule about white spacing rather than all rules):
```javascript
{ property1: value1,        property2: value2 }
```
We can use json-schemas to validate the data as well as adding tests.
To compare two different configs we can use `diff` and `vimdiff`.

> > and remove the usage of lodash.
> 
> Hmm, the code will get longer, and harder-to-read with clunky map/reduce. Lodash is part of the family of list-comprehensions; which are more expressive than map/reduce, but less expressive than a generalized query language. JQuery and D3 are two more examples in Javascript, Python has list comprehensions. C# has linq. These exist because they simplify data transformation. This codebase does a lot of data transformation; it could benefit from using chained list comprehensions that lodash provides.
> 
> List comprehensions do require a learning curve to use, but they read easier than loops and maps. Demanding more list comprehensions will force better form on the data the is being manipulated: Right now there is no pattern to the data: Every step in the transformation pipeline is a free-form transformation from one type of deep nested inner objects to another: The code looks complicated only because the data structures are complicated. Moving to list comprehensions demands some flattening of data, which makes it easier to reason about, makes transformations simpler, and results in easy-to-read list comprehension chains.
> 
I've found `lodash` makes it harder for contributors to understand the code.
We also need to evaluate it if the use of `_.chain` is adequate:
https://medium.com/making-internets/why-using-chain-is-a-mistake-9bc1f80d51ba

Perhaps in a different part of the code I could see list comprehensions be more valuable. In this specific example I see making the code harder to read.

Would you mind if we went over list comprehensions at the all hands? I want to understand your arguments for it properly.

Until then, would you mind avoiding `lodash`?

> > Also rebase the branch.
> 
> I will do this after more changes are made: I am sad that this repo is not using [gitflow](https://datasift.github.io/gitflow/IntroducingGitFlow.html)

I would like to talk about this at the all hands.

armenzg:
> On subject of separating data from format: You are suggesting I insert the
> 
> ```
>             <PerfherderGraphContainer
>               title='Tp6: Amazon'
>               series={[
>                 {
>                   label: 'Firefox',
>                   seriesConfig: {
>                     frameworkId: 10,
>                     platform: 'windows7-32',
>                     option: 'pgo',
>                     project: 'mozilla-central',
>                     suite: 'raptor-tp6-amazon-firefox',
>                   },
>                 },
>                 {
>                   label: 'Chrome',
>                   seriesConfig: {
>                     frameworkId: 10,
>                     platform: 'windows7-32-nightly',
>                     option: 'opt',
>                     project: 'mozilla-central',
>                     suite: 'raptor-tp6-amazon-chrome',
>                   },
>                 },
>               ]}
>             />
> ```
> 
> directly into `index32-bit.js`?
> 
Yes

> > I've found `lodash` makes it harder for contributors to understand the code.
> > We also need to evaluate it if the use of `_.chain` is adequate:
> > https://medium.com/making-internets/why-using-chain-is-a-mistake-9bc1f80d51ba
> 
> I have no appreciation for the optimization arguments made in that post; The proposed alternatives are all variations on the same general idea; a regex-search-an-replace can translate between them. I am happy to defer to whatever variant is popular.
> 
> But, adding this link makes your argument ambiguous: Are you asking that I use `lodash/fp/flow` instead of `lodash`? Or you asking to avoid all list comprehensions?

I prefer discussing list comprehensions later because it will be easier face to face.

With regards to `lodash` vs `lodash/path/to/something` it helps to keep the build size smaller.
The first will import all modules for lodash.
There's also published packages that are subsets of lodash to help keep the footprint small.

armenzg:
Hi @klahnakoski could you start a new view (see src/views) for this?
Use the DashboardPage like Android uses.
We can also add the quantum-new/ route.

That way we can have easier conversations around how to write things properly. The old code is very hard to change it.

Thanks!

PR #231, opened 2018-12-10
Preparation work for Js perf dashboard

armenzg:


src/components/DashboardHeader/index.jsx
```diff
@@ -0,0 +1,35 @@
...
+  },
+  subtitle: {
+      color: '#d1d2d3',
+      fontWeight: '400',
+      margin: '.5rem .6rem .5rem',
+      padding: '.5em 0',
+  },
+};
+
+const DashboardHeader = ({ classes, title, subtitle }) => (
```
klahnakoski: Why does DashboardHeader exist?  Is there a plan to use it in other dashboards?  If this component has only one call point, I suggest it stay inlined.  At the very least, give it a name that indicates it is private (use underscore as prefix, like `_DashboardHeader`)

src/components/SectionHeader/index.jsx
```diff
@@ -17,10 +22,12 @@ const styles = () => ({
 });
 
 const SectionHeader = ({ classes, title, subtitle }) => (
```
klahnakoski: Is the SectionHeader used in places other than <Section>?  This should be inlined, or made private.

src/views/JsTeam/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+    static propTypes = {
+        classes: PropTypes.shape({}).isRequired,
+    };
+
+    render() {
+        const { classes } = this.props;
+        return (
+          <div>
+            <DashboardHeader title='JavaScript Performance dashboard' />
+            <div className={classes.content}>
```
klahnakoski: I assume `classes.content` will not change; why is it a `prop`?

src/views/JsTeam/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+import DashboardHeader from '../../components/DashboardHeader';
+
+const styles = ({
+    content: {
+        margin: '1rem',
+    },
+});
+
+class JsTeam extends Component {
+    static propTypes = {
```
klahnakoski: are we going with 4-space indent, or 2-space indent?  I like the former better.

armenzg:
@klahnakoski I know this will have impact on your other PRs but I want to land it so @aimenbatool will have less CSS madness to deal with when integrating the JS perf dashboard.

If I don't hear back from you by tomorrow (you should really stay away from IRC 😄 ) I will be merging it as is.

klahnakoski:
@armenzg yes, go and merge. I saw this coming.

PR #233, opened 2018-12-11
WIP: Add tp6 page

klahnakoski:


src/quantum/index-32bit.js
```diff
@@ -22,7 +22,6 @@ export default class QuantumIndex32 extends React.Component {
   }
 
   state = {
-    apzStatus: [],
```
klahnakoski: so it matches index-64bit.js

src/quantum/index-64bit.js
```diff
@@ -627,15 +523,6 @@ export default class QuantumIndex64 extends React.Component {
...
           ],
-          [
-            <TelemetryContainer
-              key='blankWindowShown'
-              id='blankWindowShown'
-              title='Blank window shown'
-              queryParams={quantum64QueryParams}
-            />,
-            <div style={{ width: '50%' }} />,
-          ],
```
klahnakoski: to match index32-bit.js

src/quantum/index-32bit.js
```diff
@@ -10,7 +10,7 @@ import TelemetryContainer from '../telemetry/graph';
 import SETTINGS from '../settings';
 import { quantum32QueryParams, flowGraphProps, statusLabels } from './constants';
 import GraphContainer from '../components/graph-container';
-import CONFIG from './config';
+import { CONFIG } from './config';
```
klahnakoski: so it matches index-64bit.js

klahnakoski:
Only the 32 bit is working right now, this is what is looks like:

image

klahnakoski:
@armenzg 

I am ready for high level feedback, to make sure I am not too far off.

1) You will see that I added to the `styles` in more than just the one page; I did this to override what the CSS file is doing. If this is not good, then I will add more `styles` to the new page to override whatever the CSS is doing wrong.
2) Only the 32bit version is done

Thank you

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -10,6 +10,11 @@ const styles = () => ({
         marginLeft: '0.2rem',
         marginBottom: -5,
     },
+      title: {
```
armenzg: Alignment with linkIcon.

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -10,6 +10,11 @@ const styles = () => ({
...
+        backgroundColor: 'black',
+        padding: '.3em',
```
armenzg: Use `rem` instead of `em`.
IIRC `rem` is based off the global size rather than the size of the containing element.
I believe this is important when increasing/decreasing the size of a page for all elements growing in a similar ratio.

src/quantum/index-32bit.js
```diff
@@ -106,111 +105,7 @@ export default class QuantumIndex32 extends React.Component {
...
-                    frameworkId: 10,
-                    platform: 'windows7-32-nightly',
-                    option: 'opt',
-                    project: 'mozilla-central',
-                    suite: 'raptor-tp6-youtube-chrome',
-                  },
-                },
-              ]}
-            />,
+            null,
```
armenzg: Why `null`?

src/quantum/tp6-64bit.js
```diff
@@ -0,0 +1,606 @@
...
+    super(props);
+    document.body.classList.add('multipage');
+    this.fetchNotes();
+  }
+
+  state = {
+    notes: {},
+  };
+
+  async fetchNotes() {
```
armenzg: You can drop `fetchNotes` completely. We are dropping the support for the Google Spreadsheet.

src/quantum/config.js
```diff
@@ -62,4 +62,40 @@ const CONFIG = {
 
 };
 
-export default CONFIG;
+const PAGES = {
+  header:
+    ['title', 'label', 'frameworkId', 'platform', 'option', 'project', 'suite', 'extraOptions'],
```
armenzg: You can probably add one more value to identify on which page to show up (e.g. 32-bit vs 64-bit pages).
We can probably call the propery 'page' or 'viewPlatform'

src/quantum/config.js
```diff
@@ -62,4 +62,40 @@ const CONFIG = {
...
+
+  data: [
+
+    ['Tp6: Facebook', 'Firefox', 10, 'windows7-32', 'pgo', 'mozilla-central', 'raptor-tp6-facebook-firefox'],
```
armenzg: If in the future you want to make these rows shorter you can use some constants like: `MC`, `WIN7` & `FX`.

src/quantum/tp6-32bit.js
```diff
@@ -0,0 +1,89 @@
...
+  },
+  chart: {
+    justifyContent: 'center',
+    padding: '1rem',
+
+  },
+
+};
+
+const TP6_32 = ({ classes }) => {
```
armenzg: You can probably make this component generic and use the `page` or `viewPlatform` idea to determine what to show.

src/quantum/tp6-32bit.js
```diff
@@ -0,0 +1,89 @@
...
+    display: 'flex',
+  },
+  column: {
+    flex: '50%',
+    alignItems: 'center',
+  },
+  chart: {
+    justifyContent: 'center',
+    padding: '1rem',
+
```
armenzg: Extra white line

src/quantum/tp6-64bit.js
```diff
@@ -0,0 +1,606 @@
...
+    try {
+      const notes = await (await fetch(`${SETTINGS.backend}/api/perf/notes`)).json();
+      this.setState({ notes });
+    } catch (e) {
+      console.warn('Failed to fetch notes.');
+    }
+  }
+
+  render() {
+    const { full } = parse(this.props.location.search);
```
armenzg: Drop support for this.

src/quantum/tp6-32bit.js
```diff
@@ -0,0 +1,89 @@
+import _ from 'lodash';
```
armenzg: Could you please do specific imports rather than `_`? (e.g. `import { chain, zipObject } from 'lodash;`)
It keeps the bundle size smaller.

src/quantum/tp6-32bit.js
```diff
@@ -0,0 +1,89 @@
...
+const TP6_32 = ({ classes }) => {
+  const allCharts = _
+    .chain(PAGES.data)
+    // ZIP HEADER WITH ROWS TO GET OBJECTS
+    .map(row => _.zipObject(PAGES.header, row))
+    // GROUP BY title
+    .groupBy(row => row.title)
+    // LOOP OVER EACH KEY/VALUE
+    .toPairs()
+    .map(([title, series], _i) => {
```
armenzg: If you're not using `_i` keep it out of the signature.

src/quantum/tp6-32bit.js
```diff
@@ -0,0 +1,89 @@
...
+  chart: {
+    justifyContent: 'center',
+    padding: '1rem',
+
+  },
+
+};
+
+const TP6_32 = ({ classes }) => {
+  const allCharts = _
```
armenzg: Make this a function and put it outside of the render method.

src/quantum/tp6-32bit.js
```diff
@@ -0,0 +1,89 @@
...
+    .chunk(2)
+    // TRANSPOSE, SO WE HAVE TWO COLUMNS
+    .unzip()
+    .value();
+
+  return (
+    <div className={classes.body}>
+      <DashboardPage title='TP6 - Page load on 32bit'>
+        <div className={classes.area}>
+          <div className={classes.column}>
```
armenzg: What you're doing here is a bit hard to understand.
I use a `flex` container you can let the page determine if we can fit 2 or 3 or even 4 items per row.

This is also what happens when not using `flex` (see the cut out on the right side):
image

src/quantum/tp6-32bit.js
```diff
@@ -0,0 +1,89 @@
...
+    // SPLIT INTO TWO
+    .chunk(2)
+    // TRANSPOSE, SO WE HAVE TWO COLUMNS
+    .unzip()
+    .value();
+
+  return (
+    <div className={classes.body}>
+      <DashboardPage title='TP6 - Page load on 32bit'>
+        <div className={classes.area}>
```
armenzg: Another way you can approach this is the following:

- Create a `PerfherderGraphs` component
- Pass the data you generate with `lodash` to it (e.g. <PerfherderGraphs graphs={yourData} />
- And you write this:
```javascript
const { yourData } = this.state;
return (
  <div className={classes.body}>
      <DashboardPage title='TP6 - Page load on 32bit'>
         <PerfherderGraphs graphs={yourData} />
      </DashboardPage>
  </div>
);
```

By doing this you will be separating the concern of handling the data and simply presenting it.

src/quantum/index-32bit.js
klahnakoski: I do not know what else to put as a placeholder: This page demands pairs of charts.

src/quantum/tp6-64bit.js
klahnakoski: Done.  This whole page is not complete.

src/quantum/config.js
klahnakoski: yes

klahnakoski: ...64 and 32

src/quantum/tp6-32bit.js
klahnakoski: I do not think we can chain when if we import separate functions.  I will try something else

src/quantum/tp6-32bit.js
klahnakoski: this was garbage from my debugging

src/quantum/tp6-32bit.js
klahnakoski: cool, but I do not know how to get flex to decide the number of columns. 

I tried regular flow layout (`display: inline`)  but that looks like garbage: As the charts load, the page is resizing too many time.

src/quantum/tp6-32bit.js
klahnakoski: moved to constructor

src/quantum/tp6-32bit.js
klahnakoski: I think this is solved by moving the list comprehension to the constructor

src/quantum/config.js
klahnakoski: yes, maybe. Something is lost when you move data from a table into code.  Anyway, this config file has changed to handle 32 and 64 bit

klahnakoski:
@armenzg  Ready for another review.  There are three unresolved conversations (1. importing `_`, 2. dynamic number of columns, and 3. the `null`).

This version uses `<Route path='/quantum/tp6/:bits' component={TP6} />` to switch based on 32 or 64 bit.  This makes url look like:

    http://localhost:5000/quantum/tp6/64

which I really do not like.  I would prefer 

    http://localhost:5000/quantum/tp6#bits=64

which is clearer, and it can be expanded to more parameters, and it is order-insensitive.  But, I do not know how that is done yet

armenzg:
> This version uses `<Route path='/quantum/tp6/:bits' component={TP6} />` to switch based on 32 or 64 bit. This makes url look like:
> 
> ```
> http://localhost:5000/quantum/tp6/64
> ```
> 
> which I really do not like. I would prefer
> 
> ```
> http://localhost:5000/quantum/tp6#bits=64
> ```
> 
> which is clearer, and it can be expanded to more parameters, and it is order-insensitive. But, I do not know how that is done yet

You can use `location.search`:
```javascript
const searchParams = new URLSearchParams(location.search);
const bits = searchParams.get('bits');
```

You can see an example in here:
https://github.com/mozilla-frontend-infra/firefox-performance-dashboard/blob/master/src/App/index.jsx#L22-L44

src/quantum/config.js
```diff
@@ -62,4 +65,89 @@ const CONFIG = {
 
 };
 
-export default CONFIG;
+
+const PLATFORMS = [
```
armenzg: Thanks for this config adjustment!

src/quantum/config.js
```diff
@@ -62,4 +65,89 @@ const CONFIG = {
...
+  {
+    browser: 'Firefox',
+    bits: '32',
+    os: 'win',
+    label: 'Firefox',
+    frameworkId: 10,
```
armenzg: I will keep reading the code but I believe we could group the perfherder attributes in an object:
```
seriesConfig: {
    frameworkId: 10,
    platform: 'windows10-64',
    option: 'pgo',
    project: 'mozilla-central',
}
```

src/quantum/config.js
```diff
@@ -62,4 +65,89 @@ const CONFIG = {
...
+    ['Chrome', 'Tp6: Facebook', 'raptor-tp6-facebook-chrome'],
+    ['Chrome', 'Tp6: Amazon', 'raptor-tp6-amazon-chrome'],
+    ['Chrome', 'Tp6: Google', 'raptor-tp6-google-chrome'],
+    ['Chrome', 'Tp6: YouTube', 'raptor-tp6-youtube-chrome'],
+    ['Chrome', 'Tp6: imdb', 'raptor-tp6-imdb-chrome'],
+    ['Chrome', 'Tp6: imgur', 'raptor-tp6-imgur-chrome'],
+    ['Chrome', 'Tp6: wikia', 'raptor-tp6-wikia-chrome'],
+    ['Chrome', 'Tp6: bing', 'raptor-tp6-bing-chrome'],
+    ['Chrome', 'Tp6: yandex', 'raptor-tp6-yandex-chrome'],
+    ['Chrome', 'Tp6: apple', 'raptor-tp6-apple-chrome'],
```
armenzg: Could you please uppercase the name of all the sites? (e.g. apple --> Apple)

src/quantum/index-32bit.js
armenzg: Ah OK. I've been using `undefined`.

src/quantum/tp6.js
```diff
@@ -0,0 +1,101 @@
...
+      allCharts: _
+        .chain(TP6_PAGES)
+        .filter(row => row.bits === bits)
+        // GROUP BY title
+        .groupBy(row => row.title)
+        // LOOP OVER EACH KEY/VALUE
+        .toPairs()
+        .map(([title, series], i) => (
+          <div key={`page_${bits}_${i}`} className={classes.chart}>
+            <PerfherderGraphContainer
```
armenzg: Please don't put UI elements withing the state.

Instead create an object that represents your UX.
Remember "Your UI is a representation of your state".

src/quantum/tp6.js
```diff
@@ -0,0 +1,101 @@
...
+
+class TP6 extends React.Component {
+  constructor(props) {
+    super(props);
+    const { classes, match } = this.props;
+    const bits = match.params.bits;
+
+    this.state = {
+      allCharts: _
+        .chain(TP6_PAGES)
```
armenzg: Move this whole logic into a function.
You can then do something like this:
```javascript
this.state = {
   allCharts: selectChartsFromConfig(TP6_PAGES, bits, 'title')
}
```

src/quantum/tp6.js
```diff
@@ -0,0 +1,101 @@
...
+  render() {
+    const { allCharts } = this.state;
+    const { classes, match } = this.props;
+    const bits = match.params.bits;
+
+    return (
+      <div className={classes.body}>
+        <DashboardPage key={bits} title={`TP6 - Page load on ${bits} bits`}>
+          <div className={classes.area}>
+            <div className={classes.column}>
```
armenzg: To make your life easier please do one column for now.
People will be fine if we don't take advantage of the whole width.
We can visit `display: flex` later. We might want to use CSS grids.
You can also chat with hassan/sclements if you want help with this.

You can do the following once you fix the state:
```javascript
{allCharts.map(([title, series], i) => (
   <div key={`page_${bits}_${i}`} className={classes.chart}>
     <PerfherderGraphContainer
        title={title}
        series={series.map({ label, seriesConfig } => ({ label, seriesConfig })))}
      />
  </div>
)}
```

src/utils/more_lodash.js
```diff
@@ -0,0 +1,21 @@
+import _ from 'lodash';
```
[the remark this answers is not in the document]
klahnakoski: @armenzg @camd   Where can I access the exports of lodash (or any other module)?  I am spending too much time trying to guess what a module exports.

For example, import `isEqual from 'lodash/isEqual'` works, while `import chain from 'lodash/chain'` does not. But `import { chain } from 'lodash/chain'` seems to work, but neither `import { mixin } from 'lodash/mixin'` nor `import mixin from 'lodash/mixin'` work.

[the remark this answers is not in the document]
klahnakoski: Also, it seems lodash is 24kb (https://lodash.com/), given the low number of customers, and given their known high quality connection, plus the fact the chart data takes longest to arrive, I am dubious that this library is worth optimizing.

src/quantum/config.js
klahnakoski: yes.  will be in next push

src/quantum/tp6.js
klahnakoski: @armenzg Are you suggesting we sacrifice visual style for better coding style? I believe layout is complicated because of the miserable state of CSS and layout on the web**.  I think we should keep the two columns; if only to be held up as an example to be replaced by something better later.  I have no problem moving to CSS Grid, but the code will not be any simpler.

@helfi92 @sarah-clements  Do you have suggestions?  One chart per "row" seems too sparse, and probably ugly.  It would be nice if there was a CSS element that can choose the right number of columns, and fill them, but I suspect a CSS solution would be inscrutable.

** I make a distinction between "layout" (the inter-component positioning of UI elements) to "styling" (a component's look). CSS has been terrible at layout, and it continues to change; whatever we use to layout the page will probably change next year.  Specifically, the years of CSS advocates complaining that `<table>` should not be used for layout, and then to have CSS grids is comical.  Comical, not just because it is the same thing by another name, but still oblivious to how layout should be done.  When the standards committees finally use what building architects and machinists have been using to position elements for centuries (software implementations are called C.A.D) then we will have a stable, and simpler, specification for layout.

src/utils/more_lodash.js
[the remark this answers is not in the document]
klahnakoski: > I think in some cases you can get rid of lodash with a bit of map & reduce.

Yes, *all* functional expressions can be rephrased in terms of `map` and `reduce`. The Boolean operators AND and OR can also be replaced with NAND; should we get rid of AND and OR too?  

Operations that are commonly used should be named and implemented.  Some examples are: filtering, groupby, joining-tables, transposing-a-grid, selecting-and-renaming-a-subset-of-columns.  These are common operations in the data management space, and exist as primitive operations there. Lodash does a reasonable job of bringing those common operations to Javascript. If we do not use Lodash, then my next option is to re-implement Lodash functionality in `utils`.  (I would be happy to do that, since Lodash is clunky in many ways that either you or I have already noticed).

If we use common functional programming methods (aka common query operators), and we flatten our data structures, the code will get simpler.  `reduce()` should not be used; it promotes complicated data structures because it can do "anything": `reduce()` in functional programming is like `goto` in structured programming: The flexibility of the operator promotes complicated code.

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -6,11 +6,16 @@ import ChartJsWrapper from '../../components/ChartJsWrapper';
...
-    },
-  });
+  linkIcon: {
+    marginLeft: '0.2rem',
+    marginBottom: -5,
+  },
+  title: {
+    backgroundColor: 'black',
+    padding: '.3rem',
+
```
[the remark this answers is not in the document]
klahnakoski: will be fixed in next push

src/components/DashboardPage/index.jsx
```diff
@@ -3,17 +3,25 @@ import PropTypes from 'prop-types';
...
+    fontSize: '2rem',
+    padding: '1rem 1rem .7rem 1rem',
+  },
   subtitle: {
-      color: '#d1d2d3',
-      fontWeight: '400',
-      margin: '.5rem .6rem .5rem',
-      padding: '.5em 0',
+    color: '#d1d2d3',
+    fontWeight: '400',
```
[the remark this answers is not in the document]
klahnakoski: will be fixed in next push

src/quantum/config.js
```diff
@@ -62,4 +65,89 @@ const CONFIG = {
...
+    ['Chrome', 'Tp6: microsoft', 'raptor-tp6-microsoft-chrome'],
+    ['Chrome', 'Tp6: reddit', 'raptor-tp6-reddit-chrome'],
+
```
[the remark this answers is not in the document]
klahnakoski: @helfi92 is there additional `lint --fix` parameters that will do this whitespace managment automatically?

src/quantum/tp6.js
```diff
@@ -0,0 +1,101 @@
...
+        // LOOP OVER EACH KEY/VALUE
+        .toPairs()
+        .map(([title, series], i) => (
+          <div key={`page_${bits}_${i}`} className={classes.chart}>
+            <PerfherderGraphContainer
+              title={title}
+              series={_
+                .chain(series)
+                .map((s) => {
+                  return {
```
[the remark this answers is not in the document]
klahnakoski: will do.

src/quantum/tp6.js
klahnakoski: @helfi92: This looks interesting, and I will try it.

src/quantum/tp6.js
klahnakoski: @helfi92 can I generate the React elements in the constructor, rather than passing them in the props?

src/utils/more_lodash.js
```diff
@@ -0,0 +1,21 @@
...
+
+// SELECT a.*, b.*
+// FROM listA a LEFT JOIN listB b ON a.propA = b.propB
+const join = (listA, propA, listB, propB) => {
+  const lookup = _
+    .chain(listB)
+    .groupBy(rowB => rowB[propB])
+    .value();
+  return _
```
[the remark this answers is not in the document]
klahnakoski: @helfi92 I am not sure what you are suggesting here.  I am not sure what this has to do with the newline, or what eslint rules you suggest I remove.

src/quantum/tp6-32bit.js
klahnakoski: resolving this because I will try a `grid` solution

src/quantum/tp6-32bit.js
klahnakoski: resolving this conversation of more of the same with @helfi92

src/quantum/config.js
klahnakoski: You may be right, but I can not see the need yet.  If we have a `seriesConfig`, then it must also include the `suite` found in the `config.js/SUITES` table.  This will complicate the SUITES table, and it will complicate the `join` of SUITES and PLATFORMS that makes up TP6_PAGES.

I suggest your desire to group the perfherder attributes (which may be legitimate) is a side effect of the shape `<PerfherderGraphContainer>` expects from its parameters.  If `<PerfherderGraphContainer>` parameters were flat, then would you find the flatness of TP6_PAGES acceptable?

The inner-object structure is more difficult for the list comprehensions to manipulate in general.  More complicated structures require more complicated code to manipulate it.  Furthermore, this data structure is particular to `<PerfherderGraphContainer>`, and not general enough, or old enough, to be treated like black box yet.  I suggest it is premature to complicate `PLATFORMS` data structure.

src/quantum/tp6.js
helfi92: I'm not sure, you could try. However, I would rather create them in a different file and import them directly. React constructors are not typically used for creating React elements.

src/quantum/config.js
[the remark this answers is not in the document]
helfi92: I believe this rule is included in `@mozilla-frontend-infra/react-lint`.

src/utils/more_lodash.js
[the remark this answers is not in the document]
helfi92: `@mozilla-frontend-infra/react-lint` is a basically a package that sits on top of `@neutrinojs/airbnb` except it has other mozilla frontend infra additions. You can remove all the rules defined in https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/b18956d0fb79be6d4c69ccc16183e1b5ae146bfc/.neutrinorc.js#L7-L73 and only use @mozilla-frontend-infra/react-lint. `.neutrinorc.js` will end up being:

```js
module.exports = {
  options: {
    root: __dirname,
  },
  use: [
    '@mozilla-frontend-infra/react-lint',
    [
      '@neutrinojs/react',
      {
        html: {
          title: 'Firefox Health Dashboard',
        },
        devtool: {
          // Enable source-maps in production
          production: 'source-map',
        },
        env: {
          BACKEND: 'https://firefox-health-backend.herokuapp.com',
        },
        // Read https://stackoverflow.com/a/36623117
        // This is the key to making React Router work with neutrino
        // Fix issue with nested routes e.g /index/garbage
        publicPath: '/',
      },
    ],
    '@neutrinojs/jest',
    [
      '@neutrinojs/copy', {
        patterns: [
          { from: 'src/static/favicon.ico', to: 'favicon.ico' },
        ],
      },
    ],
  ],
};
```

src/quantum/config.js
armenzg: I have no problem to keeping it flat.

src/utils/more_lodash.js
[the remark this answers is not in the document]
armenzg: I beg to disagree that `reduce` makes anything complicated. We both are entitled to our opinions.

> @armenzg @camd Where can I access the exports of lodash (or any other module)? I am spending too much time trying to guess what a module exports.
> 
> For example, import `isEqual from 'lodash/isEqual'` works, while `import chain from 'lodash/chain'` does not. But `import { chain } from 'lodash/chain'` seems to work, but neither `import { mixin } from 'lodash/mixin'` nor `import mixin from 'lodash/mixin'` work.

I would install the packages and look under `./node_modules/lodash`.
Sometimes the README files of the packages is sufficient to determine this.



> Also, it seems lodash is 24kb (https://lodash.com/), given the low number of customers, and given their known high quality connection, plus the fact the chart data takes longest to arrive, I am dubious that this library is worth optimizing.

It's not a matter of how big lodash is. It's a matter if we're going to keep following a standard practice (not just within @mozilla-frontend-infra but across the web). The standard practice is that we import explicitely since the world we live in is not 100% tree shakeable. Perhaps in the future we won't have to worry but right now we do.

We simply don't have to spend time evaluating the cost of being explicit or not. We simply require to always be explicit. If you want to discuss the practices we follow feel free to start a discussion outside of the PR. It will make it easier to discuss.

src/utils/more_lodash.js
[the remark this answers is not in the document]
klahnakoski: @helfi92 Can we put changes to `.neutrinorc.js` in another PR?  I would prefer to keep this one "simple" 😄

src/utils/more_lodash.js
```diff
@@ -0,0 +1,21 @@
+import _ from 'lodash';
+
+// SELECT a.*, b.*
+// FROM listA a LEFT JOIN listB b ON a.propA = b.propB
+const join = (listA, propA, listB, propB) => {
+  const lookup = _
+    .chain(listB)
```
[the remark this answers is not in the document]
klahnakoski: will be in next push

src/utils/more_lodash.js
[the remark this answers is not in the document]
klahnakoski: @helfi92 maybe you can help me with what's wrong here:

```
C:\Users\kyle\code\firefox-health-dashboard>yarn add -D @mozilla-frontend-infra/react-lint
yarn add v1.3.2
(node:15724) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.4: The platform "win32" is incompatible with this module.
info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
error @mozilla-frontend-infra/lint@1.1.1: The engine "yarn" is incompatible with this module. Expected version ">=1.10.1".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

I tried to upgrade yarn, but the same problem happens:

```
C:\Users\kyle\code\firefox-health-dashboard>npm upgrade --global yarn
C:\Users\kyle\AppData\Roaming\npm\yarn -> C:\Users\kyle\AppData\Roaming\npm\node_modules\yarn\bin\yarn.js
C:\Users\kyle\AppData\Roaming\npm\yarnpkg -> C:\Users\kyle\AppData\Roaming\npm\node_modules\yarn\bin\yarn.js
+ yarn@1.12.3
updated 1 package in 6.481s

C:\Users\kyle\code\firefox-health-dashboard>yarn add -D @mozilla-frontend-infra/react-lint
yarn add v1.3.2
(node:19720) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.4: The platform "win32" is incompatible with this module.
info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
error @mozilla-frontend-infra/lint@1.1.1: The engine "yarn" is incompatible with this module. Expected version ">=1.10.1".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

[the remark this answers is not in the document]
klahnakoski: yes, I will save the change of the `.neutrinorc.js` for another PR.

src/quantum/tp6.js
klahnakoski: @helfi92 please take a look at what I did with Grid:  It is fine, but I would have hoped it would choose the number of columns based on the window width, and not the physical form factor.  Maybe you know more.

src/utils/more_lodash.js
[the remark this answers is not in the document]
klahnakoski: @armen  I removed `_.chain` and rewrote it as `frum` so the code reads nicer.  I am open for different naming.

klahnakoski:
@armen, I am ready for another review.

src/quantum/tp6.js
```diff
@@ -0,0 +1,64 @@
...
+  chart: {
+    justifyContent: 'center',
+    padding: '1rem',
+  },
+};
+
+class TP6 extends React.Component {
+  render() {
+    const { classes, location } = this.props;
+    const params = new URLSearchParams(location.search);
```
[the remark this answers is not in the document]
klahnakoski: yea, I really do not like the logic in this file either, in a router would be better;  First we need that tool that will help select paramters and update the URL for us.

src/quantum/tp6.js
```diff
@@ -0,0 +1,64 @@
...
+    const limits = { bits: params.get('bits') };
```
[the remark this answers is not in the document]
klahnakoski: This will be replaced with `props` eventually, ounce we have a `<Navigation>` like firox-perf-dashboard https://github.com/mozilla-frontend-infra/firefox-performance-dashboard/blob/master/src/App/index.jsx#L34

src/quantum/tp6.js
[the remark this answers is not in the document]
klahnakoski: "that tool" is `<Navigation>` like in the other project.

src/utils/query_ops.js
```diff
@@ -0,0 +1,130 @@
...
+import map from 'lodash/map';
+import flatten from 'lodash/flatten';
+import lodashFilter from 'lodash/filter';
+import toPairs from 'lodash/toPairs';
+import chunk from 'lodash/chunk';
+import unzip from 'lodash/unzip';
+import sortBy from 'lodash/sortBy';
+import fromPairs from 'lodash/fromPairs';
+
+class Wrapper {
```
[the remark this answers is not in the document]
klahnakoski: Maybe, but since this is not exported, it is private.  Maybe `ArrayWrapper`, or `EvenBetterArray`?

src/utils/query_ops.js
```diff
@@ -0,0 +1,130 @@
...
+    // USE function(){} DECLARATION TO BIND this AT CALL TIME
+    Wrapper.prototype[name] = function anonymous(...args) {
+      this.list = method(this.list, ...args);
+      return this;
+    };
+  });
+};
+
+
+const frum = (list) => {
```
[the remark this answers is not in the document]
klahnakoski: Yes, I like the word `from` because it makes the chained functions read like SQL, but it is reserved. We can go back to `chain`, but that word alludes to the chaining of functions, which is about code. while the other functions (map, filter, groupby) are words that describe action on data.

Other words could be `using` or `with`.  If `frum` is a default export, then it is private, and then does the name even matter? The importer gets to decide the name:

    import magicalChainFunction from queryOps

src/quantum/tp6.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
```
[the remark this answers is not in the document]
klahnakoski: Upon moving it there, I see it imports `/quantum/config.js`.  Are cross-directory imports like this allowed?

src/quantum/tp6.js
```diff
@@ -0,0 +1,64 @@
...
+class TP6 extends React.Component {
+  render() {
+    const { classes, location } = this.props;
+    const params = new URLSearchParams(location.search);
+    const limits = { bits: params.get('bits') };
+
+    return (
+      <div className={classes.body}>
+        <DashboardPage
+          key={limits.bits}
```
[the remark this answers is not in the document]
klahnakoski: Yes, I guess it depends on if `limits.bits` is a prop, or `state`.  I imagined it as state (since I was pulling it from `location.search` at render time.  Since you suggest moving `bits` to the constructor, I see no need for the key either

klahnakoski:
@armenzg  ready for another review.  You mentioned "Make sure to use yarn add to see package.json reflected."  but I do not see any effect:  I do not think any new libraries are needed in this PR, just parts  of ones that already exist.

src/views/QuantumTP6/index.js
```diff
@@ -2,10 +2,10 @@ import React from 'react';
 import { withStyles } from '@material-ui/core';
 import PropTypes from 'prop-types';
 import Grid from '@material-ui/core/Grid';
-import { frum } from '../utils/query_ops';
-import { TP6_PAGES } from './config';
-import DashboardPage from '../components/DashboardPage/index';
-import PerfherderGraphContainer from '../containers/PerfherderGraphContainer/index';
+import frum from '../../utils/queryOps';
+import { TP6_PAGES } from '../../quantum/config';
```
armenzg: Eventually I would like the config to move directly under `srv/views`. It's fine for now.

src/views/QuantumTP6/index.js
```diff
@@ -2,10 +2,10 @@ import React from 'react';
...
+import DashboardPage from '../../components/DashboardPage/index';
```
armenzg: You need to add `index` at the end of the path. It's implicit.
```suggestion
import DashboardPage from '../../components/DashboardPage';
```

src/views/QuantumTP6/index.js
```diff
@@ -18,25 +18,31 @@ const styles = {
...
+    super(props);
+    const { location } = this.props;
     const params = new URLSearchParams(location.search);
-    const limits = { bits: params.get('bits') };
+    this.state = { bits: params.get('bits') };
+  }
+
+  render() {
+    const { classes } = this.props;
+    const state = this.state;
```
armenzg: If the linting rules were properly working you will get alerted that destructuring is prefered.
```suggestion
    const { bits } = this.state;
```

src/views/QuantumTP6/index.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import { withStyles } from '@material-ui/core';
+import PropTypes from 'prop-types';
+import Grid from '@material-ui/core/Grid';
+import frum from '../../utils/queryOps';
+import { TP6_PAGES } from '../../quantum/config';
+import DashboardPage from '../../components/DashboardPage/index';
+import PerfherderGraphContainer from '../../containers/PerfherderGraphContainer/index';
```
armenzg:
```suggestion
import PerfherderGraphContainer from '../../containers/PerfherderGraphContainer';
```

src/views/QuantumTP6/index.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  render() {
+    const { classes } = this.props;
+    const state = this.state;
+
+    return (
+      <div className={classes.body}>
+        <DashboardPage
+          key={state.bits}
```
armenzg:
```suggestion
          key={bits}
```

src/views/QuantumTP6/index.js
```diff
@@ -0,0 +1,70 @@
...
+          title={'TP6'}
+          subtitle={`Page load on ${state.bits} bits`}
```
armenzg:
```suggestion
          subtitle={`Page load on ${bits} bits`}
```

src/views/QuantumTP6/index.js
```diff
@@ -0,0 +1,70 @@
...
+        >
+          <Grid container spacing={24}>
+            {
+              frum(TP6_PAGES)
+                .filter(state)
+                .groupBy('title')
+                .map(([series, { title }]) => (
+                  <Grid item xs={6} key={`page_${title}_${state.bits}`} className={classes.chart}>
```
armenzg:
```suggestion
                  <Grid item xs={6} key={`page_${title}_${bits}`} className={classes.chart}>
```

src/views/QuantumTP6/index.js
klahnakoski: yes, in this case I need the whole `state`, as an object, to send to the `.filter()`

klahnakoski:
I did a squash and merge; I figured you would like it better than a regular merge (with all the noise it contains).  I tried to match he commit message pattern I see in the other PRs.

Issue #213, opened 2018-08-30
Enhanced data when hovering over data points


armenzg:
Add the hg source link and revision/perf delta.

<img width="405" alt="image" src="https://user-images.githubusercontent.com/44410/43648647-0964dba4-9709-11e8-8e56-f46174a889a1.png">

armenzg:
@aimenbatool right now when hovering data points we get this:
<img width="187" alt="image" src="https://user-images.githubusercontent.com/44410/50095769-35e6cc00-01e4-11e9-9770-9700af44af88.png">

Instead we would like to see more meta data.

When you visit Perfherder (follow the hyperlink icon):
<img width="153" alt="image" src="https://user-images.githubusercontent.com/44410/50095920-84946600-01e4-11e9-837f-c1c455118644.png">
you will get to this page:
https://treeherder.mozilla.org/perf.html#/graphs?series=mozilla-central,1776360,1,10&series=mozilla-central,1775979,1,10&series=mozilla-central,1776166,1,10&selected=mozilla-central,1776166,413555,667080523,10

If you click on that same data point as on my 1st screenshot on this comment you will see the following:
<img width="291" alt="image" src="https://user-images.githubusercontent.com/44410/50096027-c9200180-01e4-11e9-9270-bc48c1d5ae82.png">

For now, pick one piece of metadata you believe you could add to PerfherderGraphContainer.
For instance, the delta between the last data point and the current one.

You can also see the metadata associated to a data point [1]: (look for `663428682`):
https://treeherder.mozilla.org/api/project/mozilla-central/performance/data/?framework=10&interval=1209600&signature_id=1775979

The push ID will then lead us to:
https://treeherder.mozilla.org/api/project/mozilla-central/resultset/412032/
On that response you will see that most recent revision for that push is `336f58aeb663c01ede2a646d51d5015bf741538d` which is used when building this link:
https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=ca0f00593e38cdab54c3a990059bbf1150e77365&tochange=336f58aeb663c01ede2a646d51d5015bf741538d

Now, you may wonder where `ca0f00593e38cdab54c3a990059bbf1150e77365` comes from?
If we look at the Perfherder data [3] we will see that the last data point prior to 412032 is 411945 which leads us to:
https://treeherder.mozilla.org/api/project/mozilla-central/resultset/411945/
[...]

PR #240, opened 2018-12-25
Issue #213 - Enhance data point, add delta value

aimenbatool:
@armenzg, @sarah-clements Please review this PR. Ignore the comments, for now, those are for my own understanding, I will remove them when merging.

### DIFF
--- a/src/utils/perfherder/chartJs/perfherderFormatter.js
+++ b/src/utils/perfherder/chartJs/perfherderFormatter.js
@@ -12,8 +12,23 @@ const generateInitialOptions = (meta) => {
   return {
     reverse: higherIsBetter,
     scaleLabel: higherIsBetter ? 'Score' : 'Load time',
+    tooltips: {
+      callbacks: {
+        footer: (tooltipItems, data) => {
+          let delta = 0;
+          if (tooltipItems[0].index > 0) {
+            delta = (data.datasets[tooltipItems[0].datasetIndex].data[tooltipItems[0].index].y
+              - data.datasets[tooltipItems[0].datasetIndex].data[tooltipItems[0].index - 1].y)
+              .toFixed(2);
+            } else {
+              delta = 'n/a';
+            }
+            return `Delta: ${delta}`;
+          },
+        },
+      },
+    };
   };
-};
 
 // This function combines Perfherder series and transforms it into ChartJS formatting
 const perfherderFormatter = (series) => {
--- a/src/utils/perfherder/chartJs/perfherderFormatter.js
+++ b/src/utils/perfherder/chartJs/perfherderFormatter.js
@@ -15,13 +15,11 @@ const generateInitialOptions = (meta) => {
     tooltips: {
       callbacks: {
         footer: (tooltipItems, data) => {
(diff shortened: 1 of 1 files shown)


aimenbatool:
This is how it looks like. 

screen shot 2018-12-25 at 7 12 46 pm

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -12,6 +12,33 @@ const generateInitialOptions = (meta) => {
   return {
     reverse: higherIsBetter,
     scaleLabel: higherIsBetter ? 'Score' : 'Load time',
+    tooltips: {
+      callbacks: {
+        footer: (tooltipItems, data) => {
+          let delta = 0;
+          tooltipItems.map((tooltipItem) => {
```
sarah-clements: You don't need to map over `tooltipItems` when there will only be one item in the array (since you are selecting one data point). In this situation, just access it via index `tooltipItems[0]` and remove the arrow function.

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -12,6 +12,33 @@ const generateInitialOptions = (meta) => {
...
+            // console.log(data.datasets[tooltipItem.datasetIndex].data[tooltipItem.index].y,
+            // ' current');
+            // console.log(data.datasets[tooltipItem.datasetIndex].data[tooltipItem.index - 1].y,
+            // 'previous');
+
+            // console.log(data.datasets, 'Data.DataSet');
+            // console.log(tooltipItem.index, 'tooltipitem.index');
+            // console.log(tooltipItem.index - 1, 'previous index');
+            // condition, if != 0 ?
```
sarah-clements: I realize this is a work in progress since there are more data points to add, but it makes it easier for the reviewer to have less comments/clutter when reviewing.

src/utils/perfherder/chartJs/perfherderFormatter.js
aimenbatool: Updated!

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -12,8 +12,23 @@ const generateInitialOptions = (meta) => {
...
+    tooltips: {
+      callbacks: {
+        footer: (tooltipItems, data) => {
+          let delta = 0;
+          if (tooltipItems[0].index > 0) {
+            delta = (data.datasets[tooltipItems[0].datasetIndex].data[tooltipItems[0].index].y
+              - data.datasets[tooltipItems[0].datasetIndex].data[tooltipItems[0].index - 1].y)
+              .toFixed(2);
+            } else {
+              delta = 'n/a';
```
sarah-clements: Something to think about for future code, is whether you actually need an `else statement`. If you wrote`let delta = 'n/a'` on line 18 instead and then only reassigned it with the `if` condition then the else isn't necessary. This is a way to write conditions in a more concise way but it doesn't apply to every situation.

PR #241, opened 2019-01-02
Issue #204 - Add tp6 page

klahnakoski:
#204 - Update dashboard for new tp6 suites (part 1)

Add QuantumTP6 view to show known tp6 metrics

* attempt to use ChartJS

* remove tp6, for moving to other page

* revert

* first version attempt

* first working version

* try this

* slightly better, prepare for url parameters

* do not need 64 bit, adding it as url parameter later

* move back to original position

* combine 32 and 64

* lint fix

* remove warning

* whitespace

* comments

* some changes from feedback

* simplify tp6, replace lodash with query_ops

* use Grid

* remove line

* move constructor logic back to render()

* whitespace

* use toArray instead of value

* use url search parameters

* remove comments

* fix ugly styles from the merge

* remove not needed styles

* docs?

* docs

* remove docs

* PR feedback

* Update src/views/QuantumTP6/index.js

Co-Authored-By: klahnakoski <klahnakoski@mozilla.com>

* Update src/views/QuantumTP6/index.js

Co-Authored-By: klahnakoski <klahnakoski@mozilla.com>

### DIFF
--- a/src/components/ChartJsWrapper/index.jsx
+++ b/src/components/ChartJsWrapper/index.jsx
@@ -11,9 +11,12 @@ const styles = {
       width: '600px',
       // Do not let it squeeze too much and deform
       minWidth: '400px',
+      background: 'white',
     },
     title: {
-      margin: '0.1rem 0 0 0',
+      color: 'white',
+      backgroundColor: 'black',
+      padding: '.3rem .3rem .3rem .3rem',
     },
 };
 
--- a/src/components/DashboardPage/index.jsx
+++ b/src/components/DashboardPage/index.jsx
@@ -7,18 +7,22 @@ const styles = {
     display: 'flex',
     flex: '1',
     flexDirection: 'column',
+    backgroundColor: 'white',
+    padding: '2rem',
   },
   title: {
     color: 'white',
     backgroundColor: 'black',
     borderBottom: '1px solid #fff',
-    padding: '0.35rem 0 0 1rem',
+    textAlign: 'center',
+    fontWeight: 100,
+    padding: '0.4rem 0 0 0',
   },
   subtitle: {
-      color: '#d1d2d3',
-      fontWeight: '400',
-      margin: '.5rem .6rem .5rem',
-      padding: '.5em 0',
+    color: '#d1d2d3',
+    fontWeight: 400,
+    margin: '.5rem .6rem .5rem',
+    padding: '.5em 0',
   },
 };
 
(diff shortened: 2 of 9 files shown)


klahnakoski:
Can we make the linking part of another issue?  I imagine there will be discussion about where the link will go, and what will surround it to be meaningful.

klahnakoski:
..or I can do this for now:

image

armenzg:
That's fine.

Let me know when you fix the CSS issue.

klahnakoski:
@armenzg 

Sorry, what is the CSS issue?

armenzg:
For the CSS issue compare this landing page:
https://deploy-preview-241--firefox-health-dashboard.netlify.com
beside this other one
https://health.graphics

klahnakoski:
@armenzg 

thanks.  I think is ti fixed now

armenzg:
I see things improved but there's one last issue on the Quantum page:
image

This PR should help catch issues like this:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/243

klahnakoski:
> 
> 
> I see things improved but there's one last issue on the Quantum page:
> image

You see something wrong:  What is it?

If you are referring to the gap right of the "Page load (tp5)": I put it there because the other charts are not in the same suite.  Plus, the data structure that controls the layout of these charts is complicated; packing the charts will result in a large diff; which will hide the intent of the change. If this gap is to be filled, can it be done in a separate PR?  Fixing the whole layout logic will be less work.

> This PR should help catch issues like this:
> #243

#243 will catch any and all changes to the page.  It can not see gaps.  Only humans can see gaps.

armenzg:
Side by side (as reference):
image

My apologies for not being clear. I think I was referring to the left padding that each graph gets on the left side but looking at it again I believe it's fine.

Could you just add a bit of a gap between the two columns rather than looking like a continuous black row?
image

Squash the commits before landing.

Congrats on getting this out the door!

klahnakoski:
"add a bit of a gap between the two columns" can be done in another PR.  

I see three possible solutions to the gap:

1. Add a margin to `<PerherderGraphContainer>` - This is not a good idea; we may want to pack charts in the future, plus it will mess with charts placed along the edge of the page (adding margin where it is not wanted)
2. Change the chart title so no bar is seen - Which is like the original, but it is sad the layout of a dashboard dictates the layout of a chart.
3. Have `index-32bit.js`: Insert a gap between the two columns of charts - I would prefer this, but the existing layout logic is complicated, and should not belong to this PR.

I will let you merge this.

Issue #211, opened 2018-08-03
Sticky data points


armenzg:
Make the data points be sticky when clicked,
Have the hg source link and revision/perf delta be sticky when clicked.

<img width="405" alt="image" src="https://user-images.githubusercontent.com/44410/43648647-0964dba4-9709-11e8-8e56-f46174a889a1.png">

armenzg:
@asingh217 if you can't get to this we will finish it up and give you credit for the original PR:
https://github.com/mozilla-frontend-infra/firefox-performance-dashboard/pull/104

Works for you?

I hope you're doing better.

aimenbatool:
@armenzg, @sarah-clements It looks like this issue is from JS Perf Dashboard. Am I suppose to go in that project and complete it or the same thing needs to be implemented here in this project? 
If yes, then where?

armenzg:
We just need to complete it in this project.

Try to look for similar components from that PR on this project.
The components might even be named the same.

PR #242, opened 2019-01-02
 Issue #211 - Add persistent tooltip

aimenbatool:
I need some thoughts here. 

As we are continuing [Akansha's work](https://github.com/asingh217/firefox-performance-dashboard/commit/1274e01f5393b8d5c7c84778eb8b0218f151fbfd), I checked and it is working fine. 

[This piece of code looks good](https://github.com/asingh217/firefox-performance-dashboard/commit/1274e01f5393b8d5c7c84778eb8b0218f151fbfd#diff-c0ed5edad59ea667c2a52980e70e7557) and i found something very similar while searching. So i decided to go with this code and not wasting time on redoing it. 

[In a comment](https://github.com/mozilla-frontend-infra/firefox-performance-dashboard/pull/104#discussion_r231936040) @armenzg  said that we want a HOC based solution. 

So, i read about HOC and tried to implement this in this project. 

I am a little confused from now onwards, In my understanding we want 

- a function under utility to deal with persistent tooltip logic
- a hoc component under /hoc/ which will accept a wrappedComponent, that wrapped component would be a ChartJSWrapper 
- we will then pass our chartJSWrapper in our HOC to implement this logic in graph 

for example if we want perfherdergraphcontainer to be stick we will have to do something [here](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/containers/PerfherderGraphContainer/index.jsx#L46) 
somehow pass the graphwrapper in hoc, and get that logic implement on it... may be.

Please check updated files, i am a little stuck at this point.

Any thought @sarah-clements or @armenzg

### DIFF
--- a/src/utils/registerTooltip.jsx
+++ b/src/utils/registerTooltip.jsx
@@ -8,10 +8,10 @@ const handleOnClick = (_, items) => {
         // the current tooltip in view has a reference to the last active tooltip
         // eslint-disable-next-line no-underscore-dangle
         Chart.lastClickedDataPoint = items[0]._chart.tooltip._lastActive[0];
-        Chart.helpers.each(Chart.instances, (chartItem) => {
-            chartItem.update();
-        });
     }
+    Chart.helpers.each(Chart.instances, (chartItem) => {
+      chartItem.update();
+  });
 };
 
 const registerPlugin = () => {
--- a/src/utils/registerTooltip.jsx
+++ b/src/utils/registerTooltip.jsx
@@ -1,0 +1,56 @@
+import Chart from 'chart.js';
+
+const handleOnClick = (_, items) => {
+    Chart.lastClickedDataPoint = null;
+    // Check if the click happened on empty chart space or data point
+    if (items.length > 0) {
+        // A Chart element has a reference to which chart it belongs to and what
+        // the current tooltip in view has a reference to the last active tooltip
+        // eslint-disable-next-line no-underscore-dangle
+        Chart.lastClickedDataPoint = items[0]._chart.tooltip._lastActive[0];
+        Chart.helpers.each(Chart.instances, (chartItem) => {
+            chartItem.update();
+        });
+    }
+};
+
+const registerPlugin = () => {
+    Chart.plugins.register({
+        id: 'tooltip-plugin',
+        beforeRender(chart) {
+          this.pluginTooltips = [];
+          // Process only if a data point was clicked
+          if (Chart.lastClickedDataPoint) {
+            // eslint-disable-next-line no-underscore-dangle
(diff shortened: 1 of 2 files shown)


sarah-clements:
I think we should go with the simple solution for now and come back to HOC once you become more familiar with react and the code base (it's not a beginner-level concept in my opinion). @armenzg does this work for you?

aimenbatool:
I have added [this events](https://www.chartjs.org/docs/latest/general/interactions/events.html#event-option) parameter to achieve the goal. I look like a [quick and graphJS related solution](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/242/files#diff-727926e1d1d9fc6166938cd2863016caR4) but I believe this doesn't fulfill the requirement.
This only shows the tooltips upon clicking, not hovering.

Here is what we have achieved with this line of code:
- make it sticky upon clicking
- click any other data point and it will close the previous one
- click anywhere outside the data point it will close the tooltip

@armenzg @sarah-clements what are thoughts on it?

We still have the previous option, this is just to show a simpler solution i found.

aimenbatool:
@sarah-clements @armenzg 
Can you please look at the latest commit and help me get unstuck. If you deploy and notice the behavior, upon clicking, it only picks the red graph line 
screen shot 2019-01-07 at 11 41 59 pm

and even if i click on other lines it still picks the same read one. 

I dry ran the code and console.log('many things'), but i am unable to pick it. 
Can you please see what is wrong here?

armenzg:
I forgot to ask. Is this ready to be reviewed?

aimenbatool:
> I forgot to ask. Is this ready to be reviewed?

Yes, it is.

src/utils/registerTooltip.jsx
```diff
@@ -0,0 +1,56 @@
...
+
+const handleOnClick = (_, items) => {
+    Chart.lastClickedDataPoint = null;
+    // Check if the click happened on empty chart space or data point
+    if (items.length > 0) {
+        // A Chart element has a reference to which chart it belongs to and what
+        // the current tooltip in view has a reference to the last active tooltip
+        // eslint-disable-next-line no-underscore-dangle
+        Chart.lastClickedDataPoint = items[0]._chart.tooltip._lastActive[0];
+        Chart.helpers.each(Chart.instances, (chartItem) => {
```
armenzg: Could you try moving this block outside of the `if` block?

I believe it would fix the issue that a tooltip will *only* go away if you click on another data point.
It does not work if you click in the white area of the chart or outside of the chart.

aimenbatool: Yes, it's working! 
Right now when we click on white space tooltip doesn't go away unless we move the mouse.

PR #253, opened 2019-01-08
WIP Register tooltip

armenzg:


src/views/JsTeam/registerTooltip.jsx
```diff
@@ -0,0 +1,57 @@
+import Chart from 'chart.js';
+
+// XXX: TODO: handle the case when we click on the same data point
```
aimenbatool: @armenzg, you commented to handle this, but i looks like its working as expected even with the same code, without additional code. 
if we click an already opened datapoint, its keeps it opened until we click any other or empty space.

PR #255, opened 2019-01-10
Bugzilla improvements

armenzg:


klahnakoski:
I assume the `getBugsData` works.  I am happy it has been moved to where it is used, instead of its own file.

Merge this.

PR #256, opened 2019-01-10
Spinner for some graph widgets + other clean up

armenzg:
This gives ChartJsWrapper a CircularProgress spinner to let the user
know that data is being fetched.

We also fix a warning of an incorrect datetime usage:
http://momentjs.com/guides/#/warnings/js-date/

There are also some other minor look changes.

### DIFF
--- a/src/components/ChartJsWrapper/index.jsx
+++ b/src/components/ChartJsWrapper/index.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Chart from 'react-chartjs-2';
+import CircularProgress from '@material-ui/core/CircularProgress';
 import { withStyles } from '@material-ui/core/styles';
 import generateOptions from '../../utils/chartJs/generateOptions';
 
@@ -20,16 +21,24 @@ const styles = {
     },
 };
 
-const ChartJsWrapper = ({ classes, data, options, title, type = 'line' }) => (
-  <div className={classes.chartContainer}>
-    {title && <h2 className={classes.title}>{title}</h2>}
-    <Chart
-      type={type}
-      data={data}
-      height={80}
-      options={generateOptions(options)}
-    />
-  </div>
+const ChartJsWrapper = ({
+  classes, data, options, title, type, chartHeight, spinnerSize,
+}) => (
+  data ? (
+    <div className={classes.chartContainer}>
+      {title && <h2>{title}</h2>}
+      <Chart
+        type={type}
+        data={data}
+        height={chartHeight}
+        options={generateOptions(options)}
+      />
+    </div>
+  ) : (
+    <div style={{ lineHeight: spinnerSize, textAlign: 'center', width: spinnerSize }}>
+      <CircularProgress />
+    </div>
+  )
 );
 
 // The properties are to match ChartJs properties
@@ -44,7 +53,7 @@ ChartJsWrapper.propTypes = {
         callbacks: PropTypes.object,
       }),
       ticksCallback: PropTypes.func,
-    }).isRequired,
+    }),
     data: PropTypes.shape({
       datasets: PropTypes.arrayOf(
         PropTypes.shape({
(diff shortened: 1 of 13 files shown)


src/components/ChartJsWrapper/index.jsx
```diff
@@ -20,16 +21,24 @@ const styles = {
...
-    {title && <h2 className={classes.title}>{title}</h2>}
-    <Chart
-      type={type}
-      data={data}
-      height={80}
-      options={generateOptions(options)}
-    />
-  </div>
+const ChartJsWrapper = ({
+  classes, data, options, title, type, chartHeight, spinnerSize,
```
klahnakoski: this is going to get out-of-control fast; as we expose more-and-more of the ChartJs properties to the API.  I suggest we plan out these parameters, and document them.  Something like:

https://github.com/klahnakoski/MoDevMetrics/blob/dev/html/modevlib/charts/chartSchema.md

src/containers/RedashContainer/index.jsx
```diff
@@ -69,14 +69,12 @@ class RedashContainer extends Component {
...
       <div>
-        {datasets && (
-          <ChartJsWrapper
-            title={title}
-            type='line'
-            data={datasets}
-            options={options}
-          />
-        )}
+        <ChartJsWrapper
```
klahnakoski: Removing these has-data checks is wonderful.

src/utils/bugzilla/chartJs/burndownDataFormatter.js
```diff
@@ -41,7 +41,7 @@ const bugsByCreationDate = (bugs, startDate) => {
     const today = new Date();
     const todaysDate = `${today.getUTCFullYear()}-${today.getMonth() + 1}-${today.getDate()}`;
     if (lastDataPoint && lastDataPoint.x !== todaysDate) {
-        accumulatedCount.push({ x: todaysDate, y: count });
+        accumulatedCount.push({ x: new Date(todaysDate), y: count });
```
klahnakoski: Is this safe?   Javascript Dates have always been an ugly mess of timezones (or the wrong timezone, or utc, or not).  I suggest sticking with milliseconds since epoch for all time variables until they are ready for presentation to the user.

src/utils/bugzilla/chartJs/burndownDataFormatter.js
```diff
@@ -53,7 +53,7 @@ const burndownDataFormatter = (bugSeries, startDate) => {
     bugSeries.forEach(({ bugs, label }, index) => {
         const bugCountPerDay = bugsByCreationDate(bugs, startDate);
         newData.data.datasets.push({
-            ...generateLineChartStyles(SETTINGS.colors[index]),
+            ...generateDatasetStyle(SETTINGS.colors[index]),
```
klahnakoski: Again, a written standard on what properties are named what would help here; even if they are not used right now; at least they will not change for every feature we will want to add control for.

test/quantum/__snapshots__/index-64.test.jsx.snap
```diff
@@ -357,6 +357,41 @@ exports[`renders correctly 1`] = `
...
+              cx={44}
+              cy={44}
+              fill="none"
+              r={20.2}
+              strokeWidth={3.6}
+              style={Object {}}
+            />
+          </svg>
+        </div>
+      </div>
```
klahnakoski: Does this test now miss the point?  Showing just progress circles rather than content?

src/utils/bugzilla/chartJs/burndownDataFormatter.js
armenzg: This is for presentation.

test/quantum/__snapshots__/index-64.test.jsx.snap
armenzg: I have not been able to get to the point where I know how to test things once there's data.
I need few days to study a good approach on how to add tests for chart.js 

For now, this only ensures that we're aware that things are changing.

src/components/ChartJsWrapper/index.jsx
armenzg: I commented on IRC I will follow up on this today.

PR #263, opened 2019-01-14
Follow up fixes for ab59b585820103fb5e778ebc3511750621494302

armenzg:


src/quantum/index.js
```diff
@@ -530,31 +529,22 @@ export default class QuantumIndex extends React.Component {
...
-                          <a href={more}>{'more'}</a>
-                          {')'}
-                        </span>
-                      );
-                    }
-                    return null;
-
-                  })()}
+              {`${title}`}
+              {more && (
```
klahnakoski: Much nicer than what I had.

src/quantum/index.js
```diff
@@ -530,31 +529,22 @@ export default class QuantumIndex extends React.Component {
           return null;
 
         })
-        .toArray();
+        .toArray().filter(Boolean);
```
klahnakoski: What is this `Boolean`?

armenzg: It helps filter out null items:
http://www.devign.me/javascript-tip-remove-falsy-items-out-of-an-array

PR #264, opened 2019-01-14
Remove warnings from test output

armenzg:


src/telemetry/graph.js
```diff
@@ -16,14 +16,18 @@ export default class TelemetryContainer extends React.Component {
...
+    try {
+      const { graphData, telemetryUrl } = await (
+        await fetch(url)).json();
+      if (!this.graphTitleLink) {
+        return;
+      }
+      this.graphTitleLink.setAttribute('href', telemetryUrl);
+      this.graphSubtitleEl.textContent = graphData.description;
+      this.graphEvolutionsTimeline(graphData, this.graphEl);
+    } catch (error) {
```
klahnakoski: What is the point of catching this here?  It looks like you are hiding the root cause by not chaining the new exception to the old. The logging-and-error-reporting on the browser is very primitive; it looks like we get to loose information.


When you say "test output" are you referring to this project's test suite, or the firefox tests that this pages shows?

armenzg: The output from |yarn test|:

> (node:87061) UnhandledPromiseRejectionWarning: Error: only absolute urls are supported
> (node:87061) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 63)
> (node:87061) UnhandledPromiseRejectionWarning: Error: only absolute urls are supported
> (node:87061) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 64)
> (node:87061) UnhandledPromiseRejectionWarning: Error: only absolute urls are supported
> (node:87061) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 65)
> (node:87061) UnhandledPromiseRejectionWarning: Error: only absolute urls are supported
> (node:87061) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 66)
> (node:87061) UnhandledPromiseRejectionWarning: Error: only absolute urls are supported
> (node:87061) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 67)

armenzg: This is only to silence the warnings.
There's an issue to add error boundaries which we can then hook up to sentry.

PR #257, opened 2019-01-10
Reference to Netlify

armenzg:

PR #266, opened 2019-01-21
Show bug burndown for GeckoView Fenix bugs

cpeterso:
The Android dashboard's bug burndown currently shows GeckoView P1 and P2/P3 bugs. We'll never actually burn down to zero P1 bugs because we'll continue to promote P2 bugs to P1. 😄 A more useful bug query to track is the number the GeckoView bugs that block the [Fenix 1.0 release](https://github.com/mozilla-mobile/fenix/).

### DIFF
--- a/src/views/Android/index.jsx
+++ b/src/views/Android/index.jsx
@@ -18,19 +18,19 @@ class Android extends Component {
             includeBugCount
             queries={[
               {
-                text: 'Open P1 bugs',
+                text: 'Open fenix:p1 bugs',
                 parameters: {
                   product: 'GeckoView',
                   resolution: '---',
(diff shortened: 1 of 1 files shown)


klahnakoski:
image

cpeterso:
Thanks!

I see the new Android graph is now live, its X axis origin is now 1970! 😁 But I swear it worked correctly when I tested on my local machine. Your screenshot also looks correct.

Is this a backend issue from missing data, e.g. PR #257?

cpeterso:
Never mind! The burndown graph looks correct now.

klahnakoski:
Good! 

image

PR #268, opened 2019-01-23
WIP: Issue #213 - enhance datapoints

aimenbatool:
Hi @eliperelman, @sarah-clements  Please review this PR.  
[Given this comment](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/213#issuecomment-451536181) I'm have added `higher is better" or "lower is better"` indicator. 
I am unable to add the line break before **delta** by any mean. I tried using `\n` or putting each information in `div>`</div>` to style it, but I'm getting errors which are breaking everything. 
Do you have any thought on how to add a line break here? 

Screenshot: 
screen shot 2019-01-23 at 5 42 42 pm

Also, does anyone know that [how this infromation ](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/213#issuecomment-456366934)is being calculated?

### DIFF
--- a/src/utils/perfherder/chartJs/perfherderFormatter.js
+++ b/src/utils/perfherder/chartJs/perfherderFormatter.js
@@ -19,18 +19,43 @@ const generateInitialOptions = meta => {
           const tooltipData = [];
           let delta = 'n/a';
           let deltaPercentage = 'n/a';
+          let jobLink = '';
+          let hgLink = '';
+          let dataset = 'n/a';
+          let currentData = 'n/a';
           if (tooltipItems[0].index > 0) {
-            const currentData = data.datasets[tooltipItems[0].datasetIndex]
-              .data[tooltipItems[0].index].y;
-            const previousData = data.datasets[tooltipItems[0].datasetIndex]
-              .data[tooltipItems[0].index - 1].y;
+            dataset = data.datasets[tooltipItems[0].datasetIndex].data;
+
+            currentData = dataset[tooltipItems[0].index].y;
+            const previousData = dataset[tooltipItems[0].index - 1].y;
 
             delta = (currentData - previousData).toFixed(2);
             deltaPercentage = (((currentData - previousData) / previousData) * 100).toFixed(2);
+
+            const tooltipDataSet = series.data[tooltipItems[0].index];
+            // make job link against job id
+            jobLink = `https://treeherder.mozilla.org/#/jobs?repo=mozilla-central&revision=ca0f00593e38cdab54c3a990059bbf1150e77365&selectedJob=${tooltipDataSet.job_id}&group_state=expanded`;
+
+            // generate hg link
+            const pushId = tooltipDataSet.push_id;
+            const previousPushID = series.data[tooltipItems[0].index - 1].push_id;
+
+            const pushiIdLink = `https://treeherder.mozilla.org/api/project/mozilla-central/resultset/${pushId}/?format=json`;
+            const previousPushiIdLink = `https://treeherder.mozilla.org/api/project/mozilla-central/resultset/${previousPushID}/?format=json`;
+
+            // get push data against pushid
+            const pushIdData = fetchPushidData(pushiIdLink);
+            const previousPushIdData = fetchPushidData(previousPushiIdLink);
+
+            const recentRevision = pushIdData.revision;
+            const previousRevision = previousPushIdData.revision;
+
+            hgLink = `https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=${previousRevision}&tochange=${recentRevision}`;
           }
-          const indicator = `${data.datasets[tooltipItems[0].datasetIndex].data[tooltipItems[0].index].y} (${higherOrLower})`;
-          delta = `Δ ${delta} (${deltaPercentage}%)`;
-          tooltipData.push(indicator, delta);
+          const indicator = `${currentData} (${higherOrLower})`;
+          jobLink = `Job(${jobLink})`;
+          hgLink = `hg (${hgLink})`;
+          tooltipData.push(indicator, `Δ ${delta} (${deltaPercentage}%)`, jobLink, hgLink);
           return tooltipData;
         },
       },
--- a/src/utils/perfherder/chartJs/perfherderFormatter.js
+++ b/src/utils/perfherder/chartJs/perfherderFormatter.js
@@ -2,14 +2,15 @@ import { parse } from 'query-string';
 import generateDatasetStyle from '../../chartJs/generateDatasetStyle';
 import SETTINGS from '../../../settings';
 
-const dataToChartJSformat = data => data.map(({ datetime, value }) => ({
-  x: datetime,
-  y: value,
-}));
-
-const generateInitialOptions = (series) => {
+const dataToChartJSformat = data =>
+  data.map(({ datetime, value }) => ({
+    x: datetime,
+    y: value,
+  }));
+const generateInitialOptions = series => {
   const higherIsBetter = !series.meta.lower_is_better;
   const higherOrLower = higherIsBetter ? 'higher is better' : 'lower is better';
+
   return {
     reverse: higherIsBetter,
     scaleLabel: higherIsBetter ? 'Score' : 'Load time',
@@ -21,6 +22,7 @@ const generateInitialOptions = (series) => {
           let deltaPercentage = 'n/a';
           let dataset = 'n/a';
           let currentData = 'n/a';
+
           if (tooltipItems[0].index > 0) {
             dataset = data.datasets[tooltipItems[0].datasetIndex].data;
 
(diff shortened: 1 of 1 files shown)


src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -21,7 +22,7 @@ const generateInitialOptions = (meta) => {
               - data.datasets[tooltipItems[0].datasetIndex].data[tooltipItems[0].index - 1].y)
               .toFixed(2);
             }
-            return `Delta: ${delta}`;
+            return `${data.datasets[tooltipItems[0].datasetIndex].data[tooltipItems[0].index].y} (${higherOrLower}) \n Delta: ${delta}`;
```
sarah-clements: Generally if you want a line break, you need to use the `<br />` tag, although in this situation I think it'd be better to have Delta as its own string on a separate line and then wrap both strings in one `<span>` or `<div>`.

aimenbatool: @sarah-clements I tried doing this also But I was getting errors. 
Let me try this again and if errors persist then I will share the error here.

sarah-clements: If you get errors then try googling the error message or for different ways to achieve he line break :)

sarah-clements:
Aimen, you can find information online if you search for 'calculating a delta percentage'.

src/utils/perfherder/chartJs/perfherderFormatter.js
aimenbatool: The error is gone. It was not related to JSX or rendering thing. It was a chartjs thing and chartJS [doesn't allow you to render any HTML element](https://stackoverflow.com/a/42095510/5864458) from a callback function. That is why nothing was working. 
Finally Fixed! :) 
Thanks.

aimenbatool:
@sarah-clements, @eliperelman PR is ready for review. 
I have added one information and updated delta value in the recent commits. Once you approve these changes I will add more information and make commits to this PR. 

I will keep this PR opened until we add all the metadata mentioned. 
Thanks.

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -7,33 +7,43 @@ const dataToChartJSformat = data => data.map(({ datetime, value }) => ({
   y: value,
 }));
 
-const generateInitialOptions = (meta) => {
-  const higherIsBetter = (meta.lower_is_better === false);
+const generateInitialOptions = (series) => {
+  const higherIsBetter = (series.meta.lower_is_better === false);
```
eliperelman: This could be written as the negation of the property since it is already a boolean:

```js
const higherIsBetter = !series.meta.lower_is_better;
```

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -7,33 +7,43 @@ const dataToChartJSformat = data => data.map(({ datetime, value }) => ({
...
           let delta = 'n/a';
+          let deltaPercentage = 'n/a';
           if (tooltipItems[0].index > 0) {
-            delta = (data.datasets[tooltipItems[0].datasetIndex].data[tooltipItems[0].index].y
-              - data.datasets[tooltipItems[0].datasetIndex].data[tooltipItems[0].index - 1].y)
-              .toFixed(2);
-            }
-            return `Delta: ${delta}`;
-          },
+            const currentData = data.datasets[tooltipItems[0].datasetIndex]
```
eliperelman: The expression `data.datasets[tooltipItems[0].datasetIndex].data` is used multiple times here, so I would extract into a variable to improve re-use and and readability:

```js
const dataset = data.datasets[tooltipItems[0].datasetIndex].data;
// ...
const currentData = dataset[tooltipItems[0].index].y
const previousData = dataset[tooltipItems[0].index - 1].y;

// ...

const indicator = `${currentData} (${higherOrLower})`;
```

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -7,33 +7,43 @@ const dataToChartJSformat = data => data.map(({ datetime, value }) => ({
...
+              .data[tooltipItems[0].index].y;
+            const previousData = data.datasets[tooltipItems[0].datasetIndex]
+              .data[tooltipItems[0].index - 1].y;
+
+            delta = (currentData - previousData).toFixed(2);
+            deltaPercentage = (((currentData - previousData) / previousData) * 100).toFixed(2);
+          }
+          const indicator = `${data.datasets[tooltipItems[0].datasetIndex].data[tooltipItems[0].index].y} (${higherOrLower})`;
+          delta = `Δ ${delta} (${deltaPercentage}%)`;
```
eliperelman: This variable doesn't need to be overwritten; it can be passed directly to the `push` call:

```js
tooltipData.push(indicator, `Δ ${delta} (${deltaPercentage}%)`);
```

aimenbatool:
Hi @eliperelman, I have made the requested changes and added a few more things to the code.
A bit of background: 
Based on [this comment](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/213#issuecomment-451536181), after adding delta and higher or lower indicator, I have now calculated job and hg link. 
The process of calculating the information is [mentioned here](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/213#issuecomment-447908593).

I am having an issue with generating hg link, I tried a few things and now I need your suggestion or a better way of solving it. 

I [created a function](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/268/files#diff-3509d2e25cacdc0cf5676f597591da6bR57) to fetch data from the given url which is not working inside the tooltip [callback function](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/268/files#diff-3509d2e25cacdc0cf5676f597591da6bL16).

Problem:
- When I use async on [callback](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/268/files#diff-3509d2e25cacdc0cf5676f597591da6bR27) (`footer: async (tooltipItems, data) => {`)  and await while caling it [here](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/268/files#diff-3509d2e25cacdc0cf5676f597591da6bR57) like ` const pushIdData = await fetchPushidData(pushiIdLink); `

It returns the data as expected and i can see that doing `console.log()`
but it doesn't appear on tooltip and makes it 
screen shot 2019-01-25 at 11 55 37 pm

- When I don't use async await or simply try to get data (which is not a valid way, but just to get other things working) it doesn't break the whole array of data, and displays remaining data perfectly like delta, indicator and job, but doesn't gives right url of hg which is obviousy becuase the data returned from function is undefined now. I have pushed the code with this 2nd case, everything else is working except hg. 

PS: I am returning the link right now for the sake of seeing the information, I will wrap this link around the text job and hg.

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -7,33 +7,78 @@ const dataToChartJSformat = data => data.map(({ datetime, value }) => ({
...
+            // generate hg link
+            const pushId = tooltipDataSet.push_id;
+            const previousPushID = series.data[tooltipItems[0].index - 1].push_id;
+
+            const pushiIdLink = `https://treeherder.mozilla.org/api/project/mozilla-central/resultset/${pushId}/?format=json`;
+            const previousPushiIdLink = `https://treeherder.mozilla.org/api/project/mozilla-central/resultset/${previousPushID}/?format=json`;
+
+            // get push data against pushid
+            const pushIdData = fetchPushidData(pushiIdLink);
+            const previousPushIdData = fetchPushidData(previousPushiIdLink);
```
sarah-clements: I think you want to do here is wrap both of these calls in `await Promise.all` so that both are executed in parallel but the code after line 58 isn't executed until those responses have returned.

aimenbatool: Hi @sarah-clements, thanks for suggesting, it works better but the actual problem is still there. 

>> It returns the data as expected and i can see that doing console.log()
but it doesn't appear on tooltip and makes it `[object promise]`
screen shot 2019-01-28 at 5 45 38 pm

I tried to find the solution for this over internet but it looks like i am missing some concept here. 
I'm unable to find any good reason for this. Any thoughts?

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -7,33 +7,78 @@ const dataToChartJSformat = data => data.map(({ datetime, value }) => ({
...
+
+            const recentRevision = pushIdData.revision;
+            const previousRevision = previousPushIdData.revision;
```
aimenbatool: When I console.log({recentRevision, previousRevision}) it gives me required results which means that we have data in variables but over tooltip, it looks like that promise isn't resolved. It is confusing...

sarah-clements: I think the easiest solution would be to modify this `/performance/data` endpoint to return the revision hash so that you don't need to perform these fetches to begin with. Let me work on that (the endpoint belongs to the Treeherder codebase) and I'll let you know when the change has landed. It will likely be a few days since it needs to be reviewed and merged.

djmitche:
It's still a work in progress -- those `undefined` values are currently only available after a fetch, and ChartJS does not allow async functions in its footer callbacks.

sarah-clements:
@aimenbatool The change to the performance/data API is live so you can remove the fetch in the tooltip and access the `revision` property on the tooltip data object.

aimenbatool:
@sarah-clements this is ready for the review.

aimenbatool:
@sarah-clements I think ekyle requested changes for undefined values when we were waiting for endpoint fix. 
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/268#pullrequestreview-198329223

sarah-clements:
@aimenbatool  Ok, I think it's fine to merge then.

PR #273, opened 2019-01-28
Fix README backend url repository

AndersonGaldin:


djmitche:
Could you submit with just the described change, and not the others (which I assume came from some automatic linting?)

PR #279, opened 2019-01-30
Issue #75 - Remove custom linting rules 

aimenbatool:

PR #281, opened 2019-01-30
Remove custom linting rules - Experimental PR (WIP)

aimenbatool:
This PR is an experimental PR. To review the original one please see
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/279

### DIFF
--- a/src/components/ChartJsWrapper/index.jsx
+++ b/src/components/ChartJsWrapper/index.jsx
@@ -7,106 +7,123 @@ import ErrorPanel from '@mozilla-frontend-infra/components/ErrorPanel';
 import generateOptions from '../../utils/chartJs/generateOptions';
 
 const styles = {
-    // This div helps with canvas size changes
-    // https://www.chartjs.org/docs/latest/general/responsive.html#important-note
-    chartContainer: {
-      width: '600px',
-      // Do not let it squeeze too much and deform
-      minWidth: '400px',
-      background: 'white',
-    },
-    title: {
-      color: 'white',
-      backgroundColor: 'black',
-      padding: '.3rem .3rem .3rem .3rem',
-    },
-    errorPanel: {
-      marginTop: '10px',
-      width: '97%',
-    },
+  // This div helps with canvas size changes
+  // https://www.chartjs.org/docs/latest/general/responsive.html#important-note
+  chartContainer: {
+    width: '600px',
+    // Do not let it squeeze too much and deform
+    minWidth: '400px',
+    background: 'white',
+  },
+  title: {
+    color: 'white',
+    backgroundColor: 'black',
+    padding: '.3rem .3rem .3rem .3rem',
+  },
+  errorPanel: {
+    marginTop: '10px',
+    width: '97%',
+  },
 };
-
 const ChartJsWrapper = ({
-  classes, data, options, title, type, chartHeight, spinnerSize,
+  classes,
+  data,
+  options,
+  title,
+  type,
+  chartHeight,
+  spinnerSize,
   missingDataError = false,
 }) => {
   let showError;
+
   if (data && missingDataError) {
-    showError = data.datasets.some((dataset) => {
+    showError = data.datasets.some(dataset => {
       const latestDataDate = new Date(dataset.data[dataset.data.length - 1].x);
       const currentDate = new Date(); // get current date
-      const timeDifference = Math.abs(currentDate.getTime() - latestDataDate.getTime());
+      const timeDifference = Math.abs(
+        currentDate.getTime() - latestDataDate.getTime()
+      );
       const daysDifference = Math.ceil(timeDifference / (1000 * 3600 * 24));
+
       return daysDifference > 3; // if days are more than 3 then show error
     });
   }
 
-  return (
-    data ? (
-      <div className={classes.chartContainer}>
-        {showError && <ErrorPanel className={classes.errorPanel} error='This item has been missing data for at least 3 days.' /> }
-        {title && <h2>{title}</h2>}
-        <Chart
-          type={type}
-          data={data}
-          height={chartHeight}
-          options={generateOptions(options)}
+  return data ? (
+    <div className={classes.chartContainer}>
+      {showError && (
+        <ErrorPanel
+          className={classes.errorPanel}
+          error="This item has been missing data for at least 3 days."
         />
-      </div>
-    ) : (
-      <div style={{ lineHeight: spinnerSize, textAlign: 'center', width: spinnerSize }}>
-        <CircularProgress />
-      </div>
-    )
+      )}
+      {title && <h2>{title}</h2>}
+      <Chart
+        type={type}
+        data={data}
+        height={chartHeight}
+        options={generateOptions(options)}
+      />
+    </div>
+  ) : (
+    <div
+      style={{
+        lineHeight: spinnerSize,
+        textAlign: 'center',
+        width: spinnerSize,
+      }}>
+      <CircularProgress />
+    </div>
   );
 };
 
 // The properties are to match ChartJs properties
 ChartJsWrapper.propTypes = {
-    classes: PropTypes.shape({}).isRequired,
-    options: PropTypes.shape({
-      reverse: PropTypes.bool,
-      scaleLabel: PropTypes.string,
-      title: PropTypes.string,
-      tooltipFormat: PropTypes.bool,
-      tooltips: PropTypes.shape({
-        callbacks: PropTypes.object,
-      }),
-      ticksCallback: PropTypes.func,
-    }),
-    data: PropTypes.shape({
-      datasets: PropTypes.arrayOf(
-        PropTypes.shape({
-          // There can be more properties than data and value,
-          // however, we mainly care about these as a minimum requirement
-          data: PropTypes.arrayOf(
-            PropTypes.shape({
-              x: PropTypes.oneOfType([
-                PropTypes.string,
-                PropTypes.instanceOf(Date),
-              ]).isRequired,
-              y: PropTypes.number.isRequired,
-            }),
-          ),
-          label: PropTypes.string.isRequired,
-        }),
-      ).isRequired,
-    }),
+  classes: PropTypes.shape({}).isRequired,
+  options: PropTypes.shape({
+    reverse: PropTypes.bool,
+    scaleLabel: PropTypes.string,
     title: PropTypes.string,
-    type: PropTypes.string,
-    chartHeight: PropTypes.number,
-    spinnerSize: PropTypes.string,
-    missingDataError: PropTypes.bool,
-    showError: PropTypes.bool,
+    tooltipFormat: PropTypes.bool,
+    tooltips: PropTypes.shape({
+      callbacks: PropTypes.object,
+    }),
+    ticksCallback: PropTypes.func,
+  }),
+  data: PropTypes.shape({
+    datasets: PropTypes.arrayOf(
+      PropTypes.shape({
+        // There can be more properties than data and value,
+        // however, we mainly care about these as a minimum requirement
+        data: PropTypes.arrayOf(
+          PropTypes.shape({
+            x: PropTypes.oneOfType([
+              PropTypes.string,
+              PropTypes.instanceOf(Date),
+            ]).isRequired,
+            y: PropTypes.number.isRequired,
+          })
+        ),
+        label: PropTypes.string.isRequired,
+      })
+    ).isRequired,
+  }),
+  title: PropTypes.string,
+  type: PropTypes.string,
+  chartHeight: PropTypes.number,
+  spinnerSize: PropTypes.string,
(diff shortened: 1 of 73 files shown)


aimenbatool:
This is an experimental PR.

sarah-clements:
Aimen can you close this if you're not going to continue using it.

aimenbatool:
@sarah-clements I would like to keep it opened for experiments until we get original one merged. 
I should add a reference to original PR in the description.

aimenbatool:
@sarah-clements PR is ready for review. Finally!!! :sleepy:

aimenbatool:
After merging the other PR this PR got more conflicts. Let me resolve this first!

aimenbatool:
@sarah-clements  It is ready for the review.

src/components/ChartJsWrapper/index.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-unresolved */
```
sarah-clements: Why is this rule being disabled here?

aimenbatool: There was some error so I disabled this rule to fix. I forgot to enable it. 
Updated!

src/components/SummaryTable/index.jsx
```diff
@@ -10,41 +11,56 @@ const SummaryTable = ({ content = [], header }) => {
...
-    </tbody>
-  </table>
+      <tbody>
+        {content.map(
+          ({ dataPoints = [], statusColor, summary, title, uid }) => (
+            <tr key={uid}>
+              <td className="title-container">
+                <StatusWidget statusColor={statusColor} title={title} />
+              </td>
+              {dataPoints.map((datum, columnIndex) => {
```
sarah-clements: Presumably the `eslint-disable react/no-array-index-key` is for this columnIndex being used as the key. Instead of disabling this rule, can you find an attribute on datum (like a name or id) that can be used instead?

src/components/genericErrorBoundary.jsx
```diff
@@ -20,12 +19,14 @@ export default class ErrorBoundary extends Component {
...
       }
+
       return (
         <div>
           <p style={{ color: 'black' }}>Oops; something went wrong</p>
         </div>
       );
     }
+
     // eslint-disable-next-line react/prop-types
```
sarah-clements: I think this should be removed and propTypes added.

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable import/no-named-as-default */
+/* eslint-disable import/no-named-as-default-member */
```
sarah-clements: I think these need to be fixed rather than disabling the rules.

src/quantum/perfherder.jsx
```diff
@@ -76,49 +107,62 @@ export default class PerfherderWidget extends React.Component {
...
+
         if (ref) {
           const refY = yScale(ref.avg);
+
           referenceYs.push(refY);
           $reference = busy ? null : (
             <line
               key={`reference-${idx}`}
-              className='reference reference-x'
+              className="reference reference-x"
```
sarah-clements: Same as earlier comment - see if you can find an attribute of the evolutions data that can be used as a key instead of the index.

src/quantum/target-status.jsx
```diff
@@ -6,6 +6,7 @@ import { statusLabels } from './constants';
 const TargetStatus = ({ notes }) => {
   const allStatus = new Map([['green', 0], ['yellow', 0], ['red', 0]]);
 
+  // eslint-disable-next-line no-restricted-syntax
```
sarah-clements: What's the default restricted syntax setting? Its something Armen might want to customize later, but it'd be good to see if there are alternatives to whats restricted in the meantime.

src/quantum/widget.jsx
```diff
@@ -4,102 +4,115 @@ import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 
-const enrich = (text, key = 'none') => (typeof text === 'string'
-  ? (
+const enrich = (text, key = 'none') =>
+  typeof text === 'string' ? (
     <span
       key={`enriched-${key}`}
+      // eslint-disable-next-line react/no-danger
```
sarah-clements: I'd find a way to replace the usage of `dangerouslySetInnerHTML` instead of disabling, or file a bug if there isn't one already to replace all uses of it.

src/quantum/widget.jsx
```diff
@@ -4,102 +4,115 @@ import { Link } from 'react-router-dom';
...
     const title = enrich(this.props.title);
     // const updated = this.props.updated && moment(this.props.updated);
-    const $title = this.props.link
-      ? (
-        <a href={this.props.link} target='_blank' rel='noopener noreferrer'>
-          {title}
-        </a>
-)
-      : title;
+    const $title = this.props.link ? (
```
sarah-clements: While you're here, can you remove the commented out code on line 29?

src/telemetry/graph.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
```
sarah-clements: Why is this rule disabled?

src/utils/BackendClient/NimbledroidApiHandler.js
```diff
@@ -1,96 +1,114 @@
...
-    }, {});
+  const mergedScenarios = productsData.reduce((result, { meta, scenarios }) => {
+    const { latestVersion, packageId } = meta;
+
+    mergedMeta[packageId] = {
+      latestVersion,
+    };
+
+    // eslint-disable-next-line consistent-return
+    Object.keys(scenarios).forEach(originalKey => {
```
sarah-clements: Please refactor this to meet the conditions of the rule.

src/quantum/index.jsx
```diff
@@ -107,14 +117,14 @@ export default class QuantumIndex extends React.Component {
         title: 'Performance Tests',
         rows: [
           <PerfherderGraphContainer
-            title='Page load (tp5)'
```
aimenbatool: @sarah-clements , @klahnakoski  
There is a linting error here in all components. 
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md 

All components which have missing `key` attribute are giving errors.  I'm not sure what should I add in `key` attribute. 

Also, is this [keys](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/7d522aaa530fd9479337f909f5ac967520570712/src/quantum/index.jsx#L60) attribute similar to the `key`? This component is also giving errors. 

What to add in key?

src/quantum/index.jsx
```diff
@@ -504,65 +513,64 @@ export default class QuantumIndex extends React.Component {
...
-        </div>
-      );
+            return null;
+          })
+          .exists()
+          .toArray();
 
-    });
+        return (
+          <div key={sectionId}>
```
aimenbatool: Here I'm getting this error. 
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

It doesn't allow us to use the array index as a key. All examples are using `.id` pattern to make it work. 
I did `{sectionId.id}` following the documentation, error removed but it didn't make sense to me. 
The same error is [here](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/quantum/index.js#L506) also. But using `.id` didnt work here.

@klahnakoski , . @sarah-clements  any thoughts how can I fix it?

src/quantum/index.jsx
djmitche: I think in this case you can add a key that corresponds to the title and is unique, so `<PerfherderGraphContainer title="Page load (tp5)" key="page-load-tp5" ...>`.  React needs the `key` property so that it can track when components move around in the virtual DOM - for example if an item is removed from an array and all of the others shift by one position.  Using a name allows React to see that the item now in position 5 in the array used to be in position 6, for example.

src/quantum/index.jsx
djmitche: In this case, this `<div>` is being returned from a `map` operation about 50 lines above:
```js
  const reduced = sections.map(
    ({ title, more, rows, cssRowExtraClasses }, sectionId) => {  ... });
```
the second argument to a map function is the index in the array, which would change if something were added or removed from the array.  That's what the lint is complaining about.

To fix, I'd suggest a similar thing to my previous suggestion: give each section a `key` property that is unique and similar to the title.  So that would mean adding a `key` property to each section in the *long* list of sections above, then passing that to the `<div>`.  Something like

```js
      ...
      {
        cssRowExtraClasses: 'generic-metrics-graphics photon-perf',
        title: 'Performance Metrics',
        key: 'performance-metrics',
        rows: [ ... ]
      }, ...
  ...
  const reduced = sections.map(
    ({ title, key, more, rows, cssRowExtraClasses }, sectionId) => {
      return <div key={key} ... >;
    });
```

src/telemetry/graph.jsx
```diff
@@ -60,16 +64,25 @@ export default class TelemetryContainer extends React.Component {
...
     return (
-      <div id={id} key={id} className='criteria-widget'>
+      <div id={id} key={id} className="criteria-widget">
         <header>
-          <h3 className='graph-title'>
-            <a className='graph-title-link' ref={a => this.graphTitleLink = a}>{title}</a>
+          <h3 className="graph-title">
+            <a
+              className="graph-title-link"
+              ref={a => (this.graphTitleLink = a)}>
```
aimenbatool: /* eslint-disable jsx-a11y/anchor-is-valid */

> Why is this rule disabled?

It has `ref` attribute in the anchor tag which makes it a clickable link. 
Linting error appears if I don't make it `href`, and if I make it an `href` attr react gives an error and breaks code. 

`ref` concept used here looks like this
https://reactjs.org/docs/refs-and-the-dom.html 

while linting error is saying that anchor isn't valid
https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md

I am confused about solving it... how can I do it?

src/components/SummaryTable/index.jsx
aimenbatool: This helped. This has been fixed now.

src/quantum/perfherder.jsx
aimenbatool: There was nothing unique in the `evolution` which could be used except the date. So i did it.

src/quantum/target-status.jsx
aimenbatool: I wasn't not sure about it. I will look into it in another PR.

src/quantum/widget.jsx
aimenbatool: done.

src/telemetry/graph.jsx
aimenbatool: @sarah-clements @djmitche your thoughts are required.

djmitche: Let's handle this in another PR.  Here's what I think is happening (and don't be confused by the similar spelling of `ref` and `href` - eslint is not confusing the two).

The link being defined here has an href which is only known after a fetch is complete (defined earlier in the same file).  The solution was to use the `ref` attribute, which is a React thing that allows you to get a reference to a DOM object, to put a reference to the link object into `this.graphTitleLink`.  Then, when the fetch is complete, there's a call to `this.graphTitleLink.setAttribute('href', telemetryUrl)` to set the href.

But eslint is complaining that this is an `a` element without an `href` -- it can't see that we plan to fill in the `href` attribute later.

I suspect the easy fix to this (and, again, let's do this in another PR!) is to put a "fake" `href` into the `a` element to start with, so that the `setAttribute` call just replaces the fake value with the real value.  I think `#` is a common "fake" href value.  So, `<a href="#" className="graph-title-link" ref={a => (this.graphTitleLink = a)}>`.

src/quantum/index.jsx
aimenbatool: I created a unique sectionId by combining sectionId + title it then worked and is now merged. 

Your suggestion is good too :) Thanks

djmitche: That sort of "cheats" by hiding the use of the array index.  But in this case, the array doesn't ever change (users can't add or remove things to it while the page is loaded), so that should be fine.  So :+1:

src/telemetry/graph.jsx
aimenbatool: @djmitche I tried doing this but it didn't work. Still gives the error. 
I tried 
href=""
href="#"
href={}
nothing worked!

It was also mentioned somewhere on a GitHub issue. It didn't work so I disabled it.

djmitche: Sounds good.  We can revisit in a dedicated PR.

Issue #276, opened 2019-01-28
Nimbledroid comparisons should compare chrome vs geckoview example


stuartphilp:
As Focus updates are not reliable/fast we should switch the Nimbledroid target comparison to be geckoviewExamplePageLoadTime / ChromeBetaPageLoadTime

I believe it's just https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/utils/nimbledroid/index.jsx#L114

stuartphilp:
We should do the cold startup graph as well (remove org.mozilla.klar and replace with reference browser, I've filed #277 for that), but what I'm referring to here is the detailed view and the individual page graphs

![screen shot 2019-01-29 at 9 02 43 am](https://user-images.githubusercontent.com/2964961/51913588-d95efb00-23a4-11e9-9220-65e0eca3b3dc.png)

"Geckoview" is the Focus/Klar app in this case, which we don't want to compare. We want Example app vs Chrome (because the Example app has nightly Geckoview builds in it so we can see the latest improvements)

klahnakoski:
finally,  https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/290

klahnakoski:
![image](https://user-images.githubusercontent.com/2334429/52285788-00817380-2935-11e9-9039-dd244468623b.png)

klahnakoski:
saw mistakes as I posted the last image:

![image](https://user-images.githubusercontent.com/2334429/52286533-5d315e00-2936-11e9-83a2-b546f8e72ec7.png)

klahnakoski:
@stuartphilp please close this when you have confirmed these PRs fixed the issue.

PR #284, opened 2019-02-01
Nimble updates

klahnakoski:
Use CONFIG in more places, standardize names to those used at nimbledroid.com

### DIFF
--- a/src/components/SummaryTable/index.jsx
+++ b/src/components/SummaryTable/index.jsx
@@ -1,30 +1,42 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import StatusWidget from '../StatusWidget';
+import CONFIG from '../../utils/nimbledroid/config';
+import { frum } from '../../utils/queryOps';
 
-const SummaryTable = ({ content = [], header }) => (
+const SummaryTable = ({ content = [], header }) => {
+  const compareName = CONFIG.packageIdLabels[CONFIG.baseProduct];
+  const compareColumn = frum(header)
+    .map((name, i) => (name === compareName ? i : null))
+    .exists()
+    .first(); // TODO: replace with findIndex()
+return (
   <table className='summary-table'>
     {header && (
-      <thead>
-        <tr>
-          <th />
-          {header.map(item => <th className='column' key={item}>{item}</th>)}
-        </tr>
-      </thead>
-    )}
+    <thead>
+      <tr>
+        <th />
+        {header.map(item => <th className='column' key={item}>{item}</th>)}
+      </tr>
+    </thead>
+      )}
     <tbody>
       {content.map(({ dataPoints = [], statusColor, summary, title, uid }) => (
         <tr key={uid}>
           <td className='title-container'>
             <StatusWidget statusColor={statusColor} title={title} />
           </td>
-          {dataPoints.map((datum, index) => <td key={index}>{datum}</td>)}
+          {dataPoints.map((datum, columnIndex) => {
+              const className = columnIndex === compareColumn ? `status-${statusColor}` : '';
+              return (<td key={columnIndex} className={className}>{datum}</td>);
+            })}
           <td>{summary}</td>
         </tr>
-            ))}
+              ))}
     </tbody>
   </table>
-);
+  );
+};
 
 SummaryTable.propTypes = ({
   content: PropTypes.arrayOf(PropTypes.shape({
--- a/src/utils/nimbledroid/index.jsx
+++ b/src/utils/nimbledroid/index.jsx
@@ -120,8 +120,11 @@ export const generateSitesTableContent = (
     yellow: 0,
     green: 0,
   };
-  const tableHeader = packageIds.map(packageId => CONFIG.packageIdLabels[packageId]);
-  tableHeader.push('% from target');
+
+  const { packageIdLabels } = CONFIG;
+  const tableHeader = packageIds.map(packageId => packageIdLabels[packageId]);
+  tableHeader.push(`% from ${packageIdLabels[baseProduct]}`);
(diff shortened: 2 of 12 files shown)


klahnakoski:
prelude to https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/276

klahnakoski:
@eliperelman please review this PR

src/components/SummaryTable/index.jsx
```diff
@@ -1,30 +1,42 @@
...
+      )}
     <tbody>
       {content.map(({ dataPoints = [], statusColor, summary, title, uid }) => (
         <tr key={uid}>
           <td className='title-container'>
             <StatusWidget statusColor={statusColor} title={title} />
           </td>
-          {dataPoints.map((datum, index) => <td key={index}>{datum}</td>)}
+          {dataPoints.map((datum, columnIndex) => {
+              const className = columnIndex === compareColumn ? `status-${statusColor}` : '';
```
sarah-clements: Could also put this in a function if you wanted to keep it single-line

src/utils/nimbledroid/index.jsx
```diff
@@ -120,8 +120,10 @@ export const generateSitesTableContent = (
     yellow: 0,
     green: 0,
   };
+
   const tableHeader = packageIds.map(packageId => CONFIG.packageIdLabels[packageId]);
-  tableHeader.push('% from target');
+  tableHeader.push(`% from ${CONFIG.packageIdLabels[CONFIG.baseProduct]}`);
+
```
sarah-clements: Rather than typing `CONFIG.packageIdLabels` constantly you could do `const { baseProduct, packageIdLabels} = CONFIG;`

src/utils/queryOps.js
```diff
@@ -22,6 +50,31 @@ class Wrapper {
...
+  index(column) {
+    // Return object indexed on column
+    // We assume the key is unique
+    const output = {};
+    this.list.forEach((row) => {
+      const key = row[column];
+      const value = output[key];
+      if (value === undefined) {
+        output[key] = row;
+      } else {
```
sarah-clements: I think you could simplify this with:
```
if (!key in output) {
  output[key] = row
}
```

src/views/QuantumTP6/index.js
```diff
@@ -39,7 +39,7 @@ class TP6 extends React.Component {
           <Grid container spacing={24}>
             {
               frum(TP6_PAGES)
-                .filter({ bits: bits })
+                .where({ bits: bits })
```
sarah-clements: Can use shorthand `{ bits }`

src/utils/queryOps.js
```diff
@@ -136,11 +191,35 @@ extend_wrapper({
...
+      };
+    }
+    return lodashFilter(list, func);
+
+  },
+
+  append: function append(list, value) {
+    return [...list, value];
+  },
+
```
sarah-clements: I'm wondering the utility of putting so many of these simple methods and syntax into a class, rather than using stand-alone helpers or just doing them in place.

src/utils/queryOps.js
```diff
@@ -136,11 +191,35 @@ extend_wrapper({
...
+    let func = null;
+
+    if (columns == null) {
+      func = row => row != null;
+    } else {
+      const cols = toArray(columns);
+      func = (row) => {
+        for (const name of cols) {
+          const v = row[name];
+          if (v == null || Number.isNaN(v)) return false;
```
sarah-clements: I think this whole block can be simplified, especially this boolean logic

src/components/SummaryTable/index.jsx
```diff
@@ -1,30 +1,42 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import StatusWidget from '../StatusWidget';
+import CONFIG from '../../utils/nimbledroid/config';
+import { frum } from '../../utils/queryOps';
```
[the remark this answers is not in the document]
klahnakoski: It is meant to reference the `FROM` clause in SQL.  I can think of no other name. I am open to suggestions.

klahnakoski:
@eliperelman nm, I got a review from @sarah-clements

klahnakoski:
@sarah-clements  thank you for the review. I either made your suggestions, or added a comment for the next PR to pickup.

klahnakoski:
@sarah-clements  I will attempt to replace `lodashToPairs` in the next PR, but it serves a useful purpose of paving over Javascript idiosyncrasies. For example, this codebase uses `Map`, which behaves differently from `Object`. Arrays, generators and iterators are also subtly different in frustrating ways.

src/utils/queryOps.js
sarah-clements: You could also do this: 
```
const transformedData = myArray.where({bits});
const newData = [...existing, ...v];
```

src/components/SummaryTable/index.jsx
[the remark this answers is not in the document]
sarah-clements: When it comes to naming, I find it helpful to think of describing what a function does (or what a variable is/contains), even if it seems overly simplistic or obvious. In my opinion, the more descriptive, the more readable -  and the less need for explanatory comments. So `frum` could be renamed to `newInstanceFrom` or `newWrapperFrom`. Though to be honest, you need this method since all its doing is adding a layer of syntax to `new Wrapper` and not adding any customized behavior :)

src/utils/queryOps.js
sarah-clements: I see what you mean about chaining methods, but you can still use the built in javascript methods without abstracting them into a class method (like .reverse()).

PR #291, opened 2019-02-05
Fix query

klahnakoski:


klahnakoski:
image

klahnakoski:
@sarah-clements  it is probably easier to review the `queryOps.js` file directly, rather than it's diff.  What's important is the simplification of the other files.

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/5613ec16d5c1f7e4b8e332f6ccf920b6c05a74e3/src/utils/queryOps.js

wlach:
I'm not sure if I know enough about the context to provide useful feedback here, sorry. :(

sarah-clements:
> I'm not sure if I know enough about the context to provide useful feedback here, sorry. :(

I'm sure I don't have enough context either. Eli is away until next Tuesday, but maybe he's a better person to review this.

helfi92:
I would typically avoid modifying existent methods of a third party library. If something is not working properly (performance issue), it can be hard to identify whether the bug is coming from the third party library or our own code. Adding new methods to a third party library on the other hand should be okay imo but I still like to create a separate utility file with all of these new methods to make it more explicit that it's not coming from somewhere else.

klahnakoski:
@helfi92 Good day!  I had forgotten Armen mentioned you can perform reviews too:  Here is a better description of what's happening:

## Chaining vs functional vs mixing 

It is sad that `chain()` is expensive to import. Lodash works around the size problem a couple of ways: First, every operation is split into its own file. Second, the `lodash/fp` flips the argument order so, instead of 

    let X = chain(A)
      .filter(B)
      .groupBy(C)
      .map(D)
      .value()

you can write

    let X = map(D, groupBy(C, filter(B, A)))

This reversed form is less elegant as function chains get longer; chaining allows one-operation-per-line, and this functional form will get messy if you have an aggressive linter:

    const X = map(
      D,
      groupBy(
        C,
        filter(
          B,
          A,
        ),
      ),
    );

You can import even less Lodash, and use `Array` functions:

    let X = groupBy(C, A.filter(B)).map(D)
 
which is even worse, since the semantic order and the logical order are now different. That is unacceptable; programs should be easy to read, and in order, as much as possible.  Programs should not be fun puzzles for you to decode.

The approach I went with was to wrap the array in a `Wrapper` class, and add extension methods for only the operations I actually use:

    let X = frum(A)
      .filter(B)
      .groupBy(C)
      .map(D)

This reads a little like SQL, which I consider a superior to ES6 in many ways. It is good that the semantic order and the logical order are congruent. But, it is bad that we have `A === frum(A)` boilerplate that the pure-functional form avoids. 

I consider the boilerplate acceptable considering the code reads in order: The boilerplate only frames the main logic, and does not weave itself into the main logic, so it is a small price to pay.

klahnakoski:
Here is some background:


# Don't import Lodash?

## The argument

I found it difficult to understand why `import _ from 'lodash';` got strong rejection from my peers:

* https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/233#discussion_r241036252
* https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/233#discussion_r241915232
* https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/233#discussion_r242160214

The main argument they gave was

> only import what you need, so that "tree shaking" can minimize the size of the executable

But this logic was not a good explanation for me:

1. I **am** including only what I needed: specifically, `chain()`, which demands all of Lodash get imported.
2. Without `chain()` I must implement the same functionality myself, which is more code, and probably slower code. 
3. Looking at [the Lodash site](https://lodash.com/), it was "only" 24kb to download. It seemed to be a small library.
4. Lodash is already included, via dependencies, over 40 times, what's the chance Lodash is not already included?
5. `import Grid from '@material-ui/core/Grid'` is just fine ([link](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/233#discussion_r242165492)) - If size is an issue, then we should be looking at each import and how it affects the size of the final executable: Importing "only what we need" from one library may be much larger than importing everything from another library.

## Analysis

Then let's measure!

> Please note: All directory listing are partial listings; they only include the relevant files.

Here is my branch, with no Lodash (https://github.com/mozilla-frontend-infra/firefox-health-dashboard/tree/2efe48fe8d6a8759fccfde70b02578383f765486)

<pre>
 Directory of C:\Users\kyle\code\firefox-health-dashboard\build\assets

2018-12-19  09:14    &lt;DIR>          .
2018-12-19  09:14    &lt;DIR>          ..
2018-12-19  09:14            20,558 1.57c6c4a5.css
2018-12-19  09:14            26,925 1.57c6c4a5.css.map
2018-12-19  09:14         1,394,782 1.5bfd5b75.js
2018-12-19  09:14         6,340,128 1.5bfd5b75.js.map
2018-12-19  09:14            80,305 index.c6516772.js
2018-12-19  09:14           231,843 index.c6516772.js.map
2018-12-19  09:14            40,410 index.ceb1cad6.css
2018-12-19  09:14            51,444 index.ceb1cad6.css.map
2018-12-19  09:14             1,495 runtime.1ca11f96.js
2018-12-19  09:14             7,996 runtime.1ca11f96.js.map
              34 File(s)      8,638,330 bytes
               2 Dir(s)  186,659,311,616 bytes free
</pre>

Here is an older version, it does the same, but using Lodash (https://github.com/mozilla-frontend-infra/firefox-health-dashboard/tree/907082170bb1b4c1b83f2ec6d84a959a89402e2b)

<pre>
 Directory of C:\Users\kyle\code\firefox-health-dashboard\build\assets

2018-12-19  09:24    &lt;DIR>          .
2018-12-19  09:24    &lt;DIR>          ..
2018-12-19  09:24            20,558 1.57c6c4a5.css
2018-12-19  09:24            26,925 1.57c6c4a5.css.map
2018-12-19  09:24         1,460,795 1.677b9c70.js
2018-12-19  09:24         6,975,948 1.677b9c70.js.map
2018-12-19  09:24            79,849 index.309df229.js
2018-12-19  09:24           230,833 index.309df229.js.map
2018-12-19  09:24            41,867 index.7572632c.css
2018-12-19  09:24            53,273 index.7572632c.css.map
2018-12-19  09:24             1,495 runtime.1ca11f96.js
2018-12-19  09:24             7,996 runtime.1ca11f96.js.map
              34 File(s)      9,341,983 bytes
               2 Dir(s)  186,649,624,576 bytes free
</pre>

This is an additional 66Kb (~5%) for Lodash. That is bigger than the download claimed.

Maybe that comparison is not fair: The two programs are not the same. Let me try the same program, with all the Lodash imports removed.  The application will not work, but at least I can see what the size differential is without Lodash:

<pre>
 Directory of C:\Users\kyle\code\firefox-health-dashboard\build\assets

2018-12-19  09:40    &lt;DIR>          .
2018-12-19  09:40    &lt;DIR>          ..
2018-12-19  09:40            20,558 1.57c6c4a5.css
2018-12-19  09:40            26,925 1.57c6c4a5.css.map
2018-12-19  09:40         1,390,059 1.d48a3844.js
2018-12-19  09:40         6,299,868 1.d48a3844.js.map
2018-12-19  09:40            41,867 index.7572632c.css
2018-12-19  09:40            53,273 index.7572632c.css.map
2018-12-19  09:40            79,234 index.841c4b52.js
2018-12-19  09:40           229,981 index.841c4b52.js.map
2018-12-19  09:40             1,495 runtime.1ca11f96.js
2018-12-19  09:40             7,996 runtime.1ca11f96.js.map
              34 File(s)      8,593,700 bytes
               2 Dir(s)  186,605,813,760 bytes free
</pre>

Confirmed.  **It seems Lodash increases the size of the app by about 70Kb (~5%)**  The "something wrong" was my understanding of the size of Lodash.  

What is the size of Lodash?

<pre>
 Directory of C:\Users\kyle\AppData\Roaming\npm\node_modules\webpack\node_modules\lodash

2017-04-19  12:27    &lt;DIR>          .
2017-04-19  12:27    &lt;DIR>          ..
2017-04-19  12:27           539,590 lodash.js
</pre>

Oh! so the "24kb download" was **very** well compressed.

## Conclusion

In conclusion, `import _ from 'lodash'` can increase the size of your executable, and maybe that is why my peers were against doing that.

It also bring up the point about tracking executable size by pull request so that we are better informed about the size of the libraries we "need" to import.

helfi92:
> let X = map(D, groupBy(C, filter(B, A)))

Would the following work?

```js
flow(
  filter(B),
  groupBy(C),
  map(D)
)(A)
```

https://lodash.com/docs/4.17.11#flow

klahnakoski:
@helfi92 

> Would the following work?
> 
> ```js
> flow(
>   filter(B),
>   groupBy(C),
>   map(D)
> )(A)
> ```

The arguments are out of order, you must import the vector functions explicitly, and you will need a `*.js` file with those vector functions defined.  

We can maintain order by using `flow2`

```js
flow2 = (list, ...args) => flow(...args)(list)
```

so we can write

> ```js
> flow2(
>   A,
>   filter(B),
>   groupBy(C),
>   map(D)
> )
> ```

Except for the additional imports required to use this form, it is trivially different that what `frum()` is doing now.

klahnakoski:
@sarah-clements 

> ...as mentioned previously, the use of 'frum' is a bit odd

I take your concern seriously, but I find naming difficult; it requires a certain artistic sense that I never had in sufficient quantity.  I am open to suggestions.  Renaming will not be hard, so we can take our time.  Also, we are discussing its mere existence, so this point may be mute.

> I haven't seen someone build their own library for performing operations on top of JS built-ins or in place of them. What's the reasoning behind it 

I have a lot of experience with SQL.  When I use other languages I see them lacking sufficiently powerful vector operations (aks list comprehensions, aka set operations, aka functional programming).  These vector operations are very useful when transforming data. Actually, it seems most code is about transforming data: Now with JSX, even displaying components is about transforming data.

Lodash has many of these vector operations, and I tried to use that library; Lodash is lacking in many ways, but at least it is well known and well supported. But, our culture does not seem to accept Lodash (see above).

So, instead, I wrote a few "utility functions" that do the common vector operations, and wrapped it in a class so I could chain them:  That is what `queryOps.js` is.

My claim is: 

> **Chained vector operations (CVOs)** makes code easier to read.  CVOs force the programmer to simplify data structures by flattening them. This flattening makes the data easier to reason about, which in turn, makes the CVOs even simpler.  

It is clear to me that my claim is not obvious.  The benefits of CVO are not seen in simple examples; rather, benefits are clearer in medium-scale software that uses complicated data structures. It turns out, the code in this repo has many examples where CVO provides benefit.

I must present more evidence to the group that CVOs are useful: I start with the nimbledroid mess:

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/300/files/c5546e894fa13b76faf7dd503c453dbc2bb62322..438c78705d526dce7206954b30908cd80fadc372?w=1

What is happening in this diff is the beginning of the code collapse: The nimbledroid transformation functions are significantly reduced, because common data transformations are done in `quuryOps.py`.  The flat datastructure is easier to inspect because object keys are replaced named properties. 

I admit this diff has some code expansion "at the edges": Where flat data must get transformed to complex structure required by the existing API.  You can see that happening [[here](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/300/files/c5546e894fa13b76faf7dd503c453dbc2bb62322..438c78705d526dce7206954b30908cd80fadc372?w=1#diff-3273e9d47a2c2378f05c49b55db54685L40)], [[here](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/300/files/c5546e894fa13b76faf7dd503c453dbc2bb62322..438c78705d526dce7206954b30908cd80fadc372?w=1#diff-6d73e9a23d47ba169f8606b30d58fd1aR116)], and in the tests [[here](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/300/files/c5546e894fa13b76faf7dd503c453dbc2bb62322..438c78705d526dce7206954b30908cd80fadc372?w=1#diff-b9c08ea53f3c0b8e18e456f49d0a2415R9)].  I hope you can see that this code will simplify when the components are made to accept the flatter data format.

This is not the end of the code collapse; This is just one step that enables more simplifications:  Some of the CVOs shown are not as simple as they could be: I deliberately made some more complicated so the diff is understandable. CVOs make the code smaller, so some functions can be combined, some files can also be combined. With CVOs more localized, more simplification will become obvious.

In the end, the nimbledroid code will have three parts: One to extract the data into flat form, one to transform the data according to some business rules, and a third to load that data into a React component for presentation. The extraction and charting can be reused; leaving one, possibly trivial, transformation for any given component.

Please notice that the CVOs are not just about simplifying code, but also about simplifying the datstructure.  What's not shown in the diff is how the complex data structure, was flattened, and useful names given to properties, and this helps with debugging. CVOs provide negligible benefit if we insist on complicated data structures; rather CVOs make simple data structures easier to use, which then makes CVOs shorter.

Now, I am not advocating rewriting existing code; but I feel https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/300 is necessary to demonstrate how CVOs are a benefit.

sarah-clements:
Thanks for the detailed explanation and for the example - I do see your point (having worked with Python, I also think it would be nice if Javascript had constructs like list comprehensions). Regarding naming - fair enough. It is tough. What do you think about `chainFrom` in place of `frum`? (see, you'll get your chain somehow :) ).

src/utils/queryOps.js
```diff
@@ -1,225 +1,416 @@
...
   }
+}
 
-  last() {
-    return last(this.list);
-  }
+function selector(columnName) {
+  // convert string into function that selects column from row
+  // convert array of strings into function that extract properties from row
+  // convert an
```
sarah-clements: Incomplete sentence?

src/utils/queryOps.js
```diff
@@ -1,225 +1,416 @@
...
-      if (value === undefined) {
-        // TODO: simplify
-        output[key] = row;
-      } else {
-        throw new Error('expecting index to be unique');
-      }
-    });
+function missing(value) {
+  // return true if value is null, or undefined, or not a legit value
+  return value == null || Number.isNaN(value) || value === '';
```
sarah-clements: You could do `!value` to cover undefined/null and ''

src/utils/queryOps.js
```diff
@@ -1,225 +1,416 @@
...
+}
 
-    return output;
-  }
+function exists(value) {
```
sarah-clements: Could you combine missing and exists?

src/utils/queryOps.js
```diff
@@ -1,225 +1,416 @@
...
 
         return true;
       };
     }
 
-    return lodashFilter(list, func);
+    return this.filter(func);
+  }
+
+  flatten() {
```
sarah-clements: Can you use `flatMap` instead?

src/utils/queryOps.js
```diff
@@ -1,225 +1,416 @@
...
+// Add Lodash functions
+extendWrapper({
+  chunk,
+  unzip,
+  zip: unzip,
+  limit: lodashTake,
+  sort: sortBy,
+  sortBy,
+
+  // SELECT a.*, b.* FROM listA a LEFT JOIN listB b on b[propB]=a[propA]
```
sarah-clements: Since this is an open-source frontend project, it'd be nice to not assume SQL knowledge have either an explanation or a commented out representation of the data structure.

src/utils/queryOps.js
klahnakoski: Zero (`0`) is a legitimate value.

src/utils/queryOps.js
klahnakoski: Maybe.  The problem I am hitting up against is the libraries (lodash) and the builtin functions are not working with Iterables in general; if the inner lists are generators, then no flattening happens; everything must be an Array.  The `for of` syntax seems to work for any Iterable, while the builtin vector functions, not so much.

That said, I have not tried flatMap, so it may work to replace this `flatten()`

src/utils/queryOps.js
klahnakoski: If it was easy to say `!missing`, then yes, but `exists` is defined as `v => !missing(v)`, which requires a bit more thought until you see it as boilerplate.

We could define a `not` functor: `not = f => ((...args) => !f(..args))` and then write `not(missing)`.  But I think this functor would have limited use, and we still have two functions.

src/utils/queryOps.js
klahnakoski: Confirmed: flatMap does not work when the inner lists are iterators.

klahnakoski:
@eliperelman Your insight would be appreciated here.

PR #292, opened 2019-02-07
fix charts when no measurements are available

klahnakoski:


src/components/NimbledroidGraph/index.jsx
```diff
@@ -9,7 +9,7 @@ class NimbledroidGraph extends Component {
   render() {
     const { profile, targetRatio } = this.props;
     const labels = CONFIG.products.map(productID => CONFIG.packageIdLabels[productID]);
-    const data = CONFIG.products.map(productID => profile.data[productID]);
+    const data = CONFIG.products.map(productID => profile.data[productID] || []);
```
klahnakoski: I tried hard to find the source of this undefined, but he nimbledroid logic was just too thick

Issue #103, opened 2018-06-12
Photon perf links can be incorrect


armenzg:
As described in here:
https://github.com/mozilla/firefox-health-dashboard/issues/99#issuecomment-396725435

Mudra01:
I would like to work on this issue. As I am new to this platform could you brief me about this a bit.

armenzg:
Sure! Have a look at https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/99#issuecomment-396725435 for more info. Load the health.graphics/quantum/64 page to see what it's meant.

Mudra01:
I checked out the page as well as the comment. I still don't get the issue and the change I need to make. Could you explain it in a bit detail.

armenzg:
Right now when we click on the header of one of the photon perf graphs we will get a link like this:
https://telemetry.mozilla.org/new-pipeline/evo.html#!aggregates=Median&arch=x86-64&cumulative=0&end_date=null&include_spill=0&keys=&max_channel_version=nightly%252F64&measure=FX_NEW_WINDOW_MS&min_channel_version=nightly%252F61&processType=*&product=Firefox&sanitize=0&sort_keys=submissions&start_date=null&trim=0&use_submission_date=0

However, the link should look like this (processType=* --> processType=parent):
https://telemetry.mozilla.org/new-pipeline/evo.html#!aggregates=Median&arch=x86-64&cumulative=0&end_date=null&include_spill=0&keys=&max_channel_version=nightly%252F64&measure=FX_NEW_WINDOW_MS&min_channel_version=nightly%252F61&processType=parent&product=Firefox&sanitize=0&sort_keys=submissions&start_date=null&trim=0&use_submission_date=0

Try to go through the code and see where that URL gets built.

aimenbatool:
@sarah-clements Can you please explain this one too?

sarah-clements:
> @sarah-clements Can you please explain this one too?

You'll need to change one of the query parameters from `processType=*` to `processType=parent` for all graph header links in the "Performance Metrics" section.

aimenbatool:
Hi @sarah-clements, @djmitche  Maybe you can add something which I am missing to fix this. 

As per this https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/99#issuecomment-395867023 and the next one of Armen it looks like the fix is on the frontend. But based on my debugging i think it is something on backend. 

[...]

PR #302, opened 2019-02-12
Issue #103 - Correct Telemetry link; overwrite default parameter

aimenbatool:
Fix for #103! 

PS: Commit typo needs to be fixed while merging the PR.

### DIFF
--- a/src/telemetry/graph.jsx
+++ b/src/telemetry/graph.jsx
@@ -26,7 +26,6 @@ export default class TelemetryContainer extends React.Component {
         return;
       }
 
-      // add processType parameter and overwrite default value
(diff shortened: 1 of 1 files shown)


src/telemetry/graph.jsx
```diff
@@ -26,7 +26,10 @@ export default class TelemetryContainer extends React.Component {
         return;
       }
 
-      this.graphTitleLink.setAttribute('href', telemetryUrl);
+      // add processType parameter and overwrite default value
```
sarah-clements: I don't think you need this comment since it is clear from looking at the code what is going on :)

aimenbatool: Removed!

PR #307, opened 2019-02-19
Fix 'n/a' values in the tooltip for initial datapoints

### LINKED ISSUE #306: Discussion: Tooltip values for initial data points
Hi everyone, I've just started tinkering with the repo in order to start contributing to it. 
I open this issue because I'm not sure if the first data points in the Perfherder graphs should display '**n/a**' in many of the tooltip fields [1].
 
I think at least `currentData` (second line from the bottom) should match the value of such point (261 at the example given). And for `delta` and `deltaPercentage`, I should suggest displaying '0' instead (there's no previous data point, so there's no increment). 
 
What do you think @aimenbatool? I think you are the one who has been working on this :sunglasses: . I have made some changes already in the code to handle this issue, but I am not sure if I should open a pull request or not. Please let me know what do you think.

[1]
screenshot_20190216_022919

2alin:
Fixes #306 

**Before:** 
screenshot_20190216_022919
**After:**
screenshot_20190218_231256

### DIFF
--- a/src/utils/perfherder/chartJs/perfherderFormatter.js
+++ b/src/utils/perfherder/chartJs/perfherderFormatter.js
@@ -17,28 +17,29 @@ const generateInitialOptions = series => {
     tooltips: {
       callbacks: {
         footer: (tooltipItems, data) => {
-          const tooltipData = [];
-          let delta = 'n/a';
-          let deltaPercentage = 'n/a';
-          let dataset = 'n/a';
-          let currentData = 'n/a';
+          // get data from all points of selected series
+          const dataset = data.datasets[tooltipItems[0].datasetIndex].data;
+          // get data from selected point
+          const currentData = dataset[tooltipItems[0].index].y;
+          const tooltipData = []; // footer's text lines will be stored here
+          let delta = 0;
+          let deltaPercentage = 0;
 
           if (tooltipItems[0].index > 0) {
-            dataset = data.datasets[tooltipItems[0].datasetIndex].data;
-
-            currentData = dataset[tooltipItems[0].index].y;
             const previousData = dataset[tooltipItems[0].index - 1].y;
 
-            delta = (currentData - previousData).toFixed(2);
-            deltaPercentage = (
-              ((currentData - previousData) / previousData) *
-              100
-            ).toFixed(2);
+            delta = currentData - previousData;
+            // [(c - p) / p] * 100 is equivalent to (c / p - 1) * 100
+            deltaPercentage = (currentData / previousData - 1) * 100;
           }
 
-          const indicator = `${currentData} (${higherOrLower})`;
+          delta = delta.toFixed(2);
+          deltaPercentage = deltaPercentage.toFixed(2);
 
-          tooltipData.push(indicator, `Δ ${delta} (${deltaPercentage}%)`);
+          const indicator = `${currentData} (${higherOrLower})`;
+          const deltaLine = `Δ ${delta} (${deltaPercentage}%)`;
+
+          tooltipData.push(indicator, deltaLine);
 
           return tooltipData;
         },
--- a/src/utils/perfherder/chartJs/perfherderFormatter.js
+++ b/src/utils/perfherder/chartJs/perfherderFormatter.js
@@ -17,24 +17,22 @@ const generateInitialOptions = series => {
     tooltips: {
       callbacks: {
         footer: (tooltipItems, data) => {
+          const tooltipData = []; // footer's text lines will be stored here
           // get data from all points of selected series
           const dataset = data.datasets[tooltipItems[0].datasetIndex].data;
           // get data from selected point
           const currentData = dataset[tooltipItems[0].index].y;
-          const tooltipData = []; // footer's text lines will be stored here
-          let delta = 0;
-          let deltaPercentage = 0;
 
           tooltipData.push(`${currentData} (${higherOrLower})`);
 
           if (tooltipItems[0].index > 0) {
             const previousData = dataset[tooltipItems[0].index - 1].y;
-
-            delta = (currentData - previousData).toFixed(2);
+            const delta = (currentData - previousData).toFixed(2);
             // [(c - p) / p] * 100 is equivalent to (c / p - 1) * 100
-            deltaPercentage = ((currentData / previousData - 1) * 100).toFixed(
-              2
-            );
+            const deltaPercentage = (
+              (currentData / previousData - 1) *
+              100
+            ).toFixed(2);
 
             tooltipData.push(`Δ ${delta} (${deltaPercentage}%)`);
           }
--- a/src/utils/perfherder/chartJs/perfherderFormatter.js
+++ b/src/utils/perfherder/chartJs/perfherderFormatter.js
@@ -25,21 +25,19 @@ const generateInitialOptions = series => {
           let delta = 0;
           let deltaPercentage = 0;
 
+          tooltipData.push(`${currentData} (${higherOrLower})`);
+
           if (tooltipItems[0].index > 0) {
             const previousData = dataset[tooltipItems[0].index - 1].y;
 
-            delta = currentData - previousData;
+            delta = (currentData - previousData).toFixed(2);
             // [(c - p) / p] * 100 is equivalent to (c / p - 1) * 100
-            deltaPercentage = (currentData / previousData - 1) * 100;
+            deltaPercentage = ((currentData / previousData - 1) * 100).toFixed(
+              2
+            );
+
+            tooltipData.push(`Δ ${delta} (${deltaPercentage}%)`);
           }
-
-          delta = delta.toFixed(2);
-          deltaPercentage = deltaPercentage.toFixed(2);
-
-          const indicator = `${currentData} (${higherOrLower})`;
-          const deltaLine = `Δ ${delta} (${deltaPercentage}%)`;
-
-          tooltipData.push(indicator, deltaLine);
 
           return tooltipData;
         },


2alin:
I usually add plenty of documentation to my code, if I need to remove some comment lines, please let me know :)

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -17,28 +17,29 @@ const generateInitialOptions = series => {
...
-              ((currentData - previousData) / previousData) *
-              100
-            ).toFixed(2);
+            delta = currentData - previousData;
+            // [(c - p) / p] * 100 is equivalent to (c / p - 1) * 100
+            deltaPercentage = (currentData / previousData - 1) * 100;
           }
 
+          delta = delta.toFixed(2);
+          deltaPercentage = deltaPercentage.toFixed(2);
```
aimenbatool: @2alin it is good to chain .toFixed(2) on line 31 and 33 it will reduce line of code and make it look more clean. 

You can do 
`delta = (currentData - previousData).toFixed(2);`
`deltaPercentage = ((currentData / previousData - 1) * 100).toFixed(2);`

aimenbatool: Also, you can remove the comment. :)

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -17,28 +17,29 @@ const generateInitialOptions = series => {
...
-            currentData = dataset[tooltipItems[0].index].y;
             const previousData = dataset[tooltipItems[0].index - 1].y;
 
-            delta = (currentData - previousData).toFixed(2);
-            deltaPercentage = (
-              ((currentData - previousData) / previousData) *
-              100
-            ).toFixed(2);
+            delta = currentData - previousData;
+            // [(c - p) / p] * 100 is equivalent to (c / p - 1) * 100
```
djmitche: These are good comments: they explain the next line differently, helping the reader understand what it is doing.

src/utils/perfherder/chartJs/perfherderFormatter.js
2alin: I moved the 'fixes' outside the conditional for a reason: to make the number format uniform even for the initial points which are not considered in the conditional. If I move it inside, as you suggest, I'll end up with Δ 0 (0%) for the initial point in comparison with all the other values like Δ 2.00 (12.13%) and even values that are not initial but end up being  Δ 0.00 (0.00%). In my opinion, it doesn't look good to have different number format for the same kind of values.

src/utils/perfherder/chartJs/perfherderFormatter.js
2alin: Thank you

2alin:
So @aimenbatool @djmitche, I'm a little confused, should I remove the comments in lines 20, 23, 24 (where variables are initialized) or keep all the comments?

src/utils/perfherder/chartJs/perfherderFormatter.js
aimenbatool: This makes sense. Let's wait for @sarah-clements  to come and review then we should be able to merge this PR.

src/utils/perfherder/chartJs/perfherderFormatter.js
djmitche: I didn't mean to contradict @aimenbatool - on balance, I think it's best to keep comments except when they are totally redundant.  I often see things like:
```js
// set x to 0
x = 0;
```
or
```js
// export DataComponent
module.exports = DataComponent;
```
and those comments are definitely worth deleting because they don't help the reader.  But your comments here help someone not familiar with the code to figure out what it means a little faster, so I think they can stay.

aimenbatool: @2alin Keep the comments :) I go with what Dustin says.

2alin: Thank you @djmitche and @aimenbatool, these little discussions are good for me because I end up understanding how you all think as team :smiley:

sarah-clements: I agree it is helpful to have the comments in this situation. Some of the best advice I've been given is to name variables and functions in a way that eliminates the need for a comment where possible.
So instead of:
```
// get data from all points of selected series
const dataset = data.datasets[tooltipItems[0].datasetIndex].data;
```

You could write:
`const allSeriesData` or `const selectedSeriesData`

Don't feel you need to change it - this is something to consider for future work. :)

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -17,28 +17,29 @@ const generateInitialOptions = series => {
...
-          let deltaPercentage = 'n/a';
-          let dataset = 'n/a';
-          let currentData = 'n/a';
+          // get data from all points of selected series
+          const dataset = data.datasets[tooltipItems[0].datasetIndex].data;
+          // get data from selected point
+          const currentData = dataset[tooltipItems[0].index].y;
+          const tooltipData = []; // footer's text lines will be stored here
+          let delta = 0;
+          let deltaPercentage = 0;
```
sarah-clements: It makes it easier to review what's changed if the order of the existing variables stays the same (unless there was a reason to change it).

src/utils/perfherder/chartJs/perfherderFormatter.js
sarah-clements: If the value is 0, I think it's fine to have the format be `Δ 0 (0%)`. I don't think it is necessarily better to show `Δ 0.00 (0.00%)` since the decimals are irrelevant.

sarah-clements:
Also, please have @klahnakoski review subsequent pull requests - he is an active contributor/maintainer of this repo.

src/utils/perfherder/chartJs/perfherderFormatter.js
2alin: not an important reason: `dataset`and `currrentData`are now 'constants' so I thought they should go at the top along with `tooltip` which is constant as well, otherwise I would end up having the next lines `const` -> `let`-> `let`->`const`->`const`.
Should I change the order then?

2alin:
Thank you all for your reviews, waiting for someone to merge it then :)

klahnakoski:
Hmm, this change has a qualitative aspect to it:  Is `Δ n/a`  better than `Δ 0.00`?  To me, the former is definitely more ugly, while the latter is misleading. `Δ 0.00` implies there is a previous point (off the chart), and it is exactly the same value as the first point shown.

Please make a new PR that does not include the `deltaLine` when a delta can not be calculated

2alin:
>  while the latter is misleading. `Δ 0.00` implies there is a previous point (off the chart), and it is exactly the same value as the first point shown.

@klahnakoski yup, you are right there. And I agree with the idea of removing deltaLine if it's the first datapoint. Give me a couple of minutes and I'll fix it in this same PR with a new commit. 

Also, this new approach allows me to implement what @aimenbatool suggested about keeping the 'toFix' inside the conditional, because we won't need delta line for the points outside of the conditional.

src/utils/perfherder/chartJs/perfherderFormatter.js
sarah-clements: I'm not sure the ordering of `const` and `let` matters - it makes it easier to read the diff to not reorder. But it's moot at this point so you can leave it.

2alin:
Done, @klahnakoski could you review it again? please.

My apologies in advance for relocating some lines of code, I know it's difficult for you to review it. I tried to find a way to keep the 'pushes' of `tooltipData` as they were before (at the end after the conditional) but the code started to get more and more complex because I needed to add an extra conditional so I could avoid adding `deltaLine` for initial points.

At the end I thought it was better to make the first push before the conditional and then the `deltaLine` push inside the conditional.

Also, I thought we could save two lines of code, adding the content of `indicator` and `deltaLine` directly into the array push method because it was being done separately now and it would be easy to understand.

klahnakoski:
@2alin starting review now.  In the meantime, I am ekyle on irc..mozilla.com, and you can find me on the #frontend-infra channel. You can ping me there for a faster response.  If you can connect, then I can send you an irc.cloud license that stays connected when your machine is not.

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -17,28 +17,27 @@ const generateInitialOptions = series => {
...
-          let delta = 'n/a';
-          let deltaPercentage = 'n/a';
-          let dataset = 'n/a';
-          let currentData = 'n/a';
+          // get data from all points of selected series
+          const dataset = data.datasets[tooltipItems[0].datasetIndex].data;
+          // get data from selected point
+          const currentData = dataset[tooltipItems[0].index].y;
+          const tooltipData = []; // footer's text lines will be stored here
+          let delta = 0;
```
klahnakoski: remove this line, use `const delta = ....` below

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -17,28 +17,27 @@ const generateInitialOptions = series => {
...
+          let deltaPercentage = 0;
```
klahnakoski: same with this line

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -17,28 +17,27 @@ const generateInitialOptions = series => {
...
-          const tooltipData = [];
-          let delta = 'n/a';
-          let deltaPercentage = 'n/a';
-          let dataset = 'n/a';
-          let currentData = 'n/a';
+          // get data from all points of selected series
+          const dataset = data.datasets[tooltipItems[0].datasetIndex].data;
+          // get data from selected point
+          const currentData = dataset[tooltipItems[0].index].y;
+          const tooltipData = []; // footer's text lines will be stored here
```
klahnakoski: make this the first line of the block, it is the complement to the line at the end of the block (`return tooltipData`)

src/utils/perfherder/chartJs/perfherderFormatter.js
2alin: Missed these two, thank you.

src/utils/perfherder/chartJs/perfherderFormatter.js
2alin: oooh this one, I love it, new for me :grin:

2alin:
Done @klahnakoski, and yup I'll ping you in IRC for any doubt I get while working with other issues. Thank you!

klahnakoski:
For next time: 

There is no need comment single lines; There are few lines that can be complicated enough to warrant a comment. I am mostly concerned with what happens when these lines get changed; there will be a tendency to keep the comment, and comments get stale.  

The math comment may be an exception. Even so, I would advocate a less "efficient" calculation that better exposes the relationship between the delta and the percentage without a comment. 

Since you were rewriting `deltaPercentage` you could also simplify it. Instead of 

```js
          if (tooltipItems[0].index > 0) {
            const previousData = dataset[tooltipItems[0].index - 1].y;
            const delta = (currentData - previousData).toFixed(2);
            // [(c - p) / p] * 100 is equivalent to (c / p - 1) * 100
            const deltaPercentage = (
              (currentData / previousData - 1) *
              100
            ).toFixed(2);

            tooltipData.push(`Δ ${delta} (${deltaPercentage}%)`);
          }
```

the `deltaPercentage` can use the `delta`, and the formatting should be delayed until the very last line:

```js
          if (tooltipItems[0].index > 0) {
            const previousData = dataset[tooltipItems[0].index - 1].y;
            const delta = currentData - previousData
            const deltaPercentage = delta / previousData *100;

            tooltipData.push(`Δ ${delta.toFixed(2)} (${deltaPercentage.toFixed(2)}%)`);
          }
```

2alin:
We had a discussion previously about the comments, some people thought they were unnecessary and we should remove them and others thought they were good for newcomers, at the end we decided to kept them. I don't mind about reducing the amount of documentation, that's why I asked at the top of this thread because I noticed that the code in this repo was very clean about them. 

And yup you are right about simplifying deltaPercentage, I'll try to remember it next time I commit to this file. :)

aimenbatool:
@2alin congratulations on your first landing. 🎉

2alin:
@aimenbatool Thank you!!

Issue #83, opened 2018-04-05
Add error handling


sarah-clements:
For all of the Quantum routes and crashes/beta (and possibly others?), if an endpoint fails the page continues to show "loading" indefinitely.

sarah-clements:
It would also be a good idea to add error handling that informs the user that they failed to enter required query parameters.

aimenbatool:
@sarah-clements Can you show how to reproduce the error to see what need to be fixed and where?
Please explain this.

sarah-clements:
[These](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/routes.jsx) are the routes I was referring to. So as an example, if I type something incorrectly [here](https://health.graphics/quantum/64/kraken), the page only shows 'loading' instead of a useful error message that my query parameters are wrong and what I need to supply. This would be for the route '/quantum/:platform/:suite' with `platform` and `suite` being query parameters (this might be different from what you've seen, ex: `platform=64`, that can be done with[ react-router](https://reacttraining.com/react-router/web/guides/philosophy)). 

With that being said, since I filed this bug almost a year ago and I don't know all that has changed within this app, it'd be worth checking with @klahnakoski and coordinating with him on what else might need useful error handling either from lack of query parameters or from API's failing. If the issue has already been solved in different ways, you can post a short explanation and close out the bug.

klahnakoski:
@aimenbatool 

The problem is best seen here https://health.graphics/quantum/64

Open the debugger and watch the network requests.  There are many telemetry requests, and some always fail.  It would be nice to have the chart space filled with some indication that it failed.

The telemetry charts are still using the old school metrics graphics and are not wrapped with any "loading" indicator. I do not suggest you replace them with ChartJS versions, instead. "just" wrap them in loading/error indicators.

On the subject of "wrapping them with loading/error indicators". I realize we do not have a function for that yet.  You will notice that we have lots of switching code. Here is one example:

```js
  return data ? (
    <div className={classes.chartContainer}>
      {showError && (
        <ErrorPanel
          className={classes.errorPanel}
          error="This item has been missing data for at least 3 days."
        />
      )}
      {title && <h2>{title}</h2>}
      <Chart
        type={type}
        data={data}
        height={chartHeight}
        options={generateOptions(options)}
      />
    </div>
  ) : (
    <div
      style={{
        lineHeight: spinnerSize,
        textAlign: 'center',
[...]

PR #309, opened 2019-02-19
Issue #83 - Add error handling

aimenbatool:
Fix for #83 
@sarah-clements I went with your suggestion using error panel to display error as I'm not comfortable with HOC yet I would try that v soon when I feel confident working with that. 

To test:
go to http://localhost:5000/quantum/64 and check for failed telemetry requests, an error message is displayed here. 

Disconnect from the internet, go to http://localhost:5000/quantum/tp6?bits=64 
You will see error msgs when there is no data. 

For this case, having the wrong parameter ( 323 instead of 32 of 64) 
https://health.graphics/quantum/323 

The code is written to handle this issue as 
if the user enters '32' then show data related to '32' bit else show '64' bit data no matter what value is entered. [See code](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/quantum/index.jsx#L36-L42)

So if the user enters 32 then it will render 32 bit data otherwise for any case it will throw 64 bit. 

Only this case is left: https://health.graphics/quantum/tp6?bits=63 
which I am figuring out to throw error its in process

Meanwhile please review the current PR.

### DIFF
--- a/src/components/ChartJsWrapper/index.jsx
+++ b/src/components/ChartJsWrapper/index.jsx
@@ -21,8 +21,8 @@ const styles = {
     padding: '.3rem .3rem .3rem .3rem',
   },
   errorPanel: {
-    marginTop: '10px',
-    width: '97%',
+    margin: '26px auto 40px',
+    width: '70%',
   },
 };
 const ChartJsWrapper = ({
@@ -73,6 +73,10 @@ const ChartJsWrapper = ({
         textAlign: 'center',
         width: spinnerSize,
       }}>
+      <ErrorPanel
+        className={classes.errorPanel}
+        error="Something went wrong, please try again later."
+      />
       <CircularProgress />
     </div>
   );
@@ -123,7 +127,7 @@ ChartJsWrapper.defaultProps = {
   title: '',
   type: 'line',
   chartHeight: 80,
-  spinnerSize: '8rem',
+  spinnerSize: '100%',
 };
 
 export default withStyles(styles)(ChartJsWrapper);
--- a/src/components/ChartJsWrapper/index.jsx
+++ b/src/components/ChartJsWrapper/index.jsx
@@ -111,6 +111,7 @@ ChartJsWrapper.propTypes = {
   }),
   title: PropTypes.string,
   type: PropTypes.string,
+  isLoading: PropTypes.bool,
   chartHeight: PropTypes.number,
   spinnerSize: PropTypes.string,
   missingDataError: PropTypes.bool,
@@ -124,6 +125,7 @@ ChartJsWrapper.defaultProps = {
   type: 'line',
   chartHeight: 80,
   spinnerSize: '100%',
+  isLoading: false,
 };
 
 export default withStyles(styles)(ChartJsWrapper);
--- a/src/components/ChartJsWrapper/index.jsx
+++ b/src/components/ChartJsWrapper/index.jsx
@@ -73,10 +73,6 @@ const ChartJsWrapper = ({
         textAlign: 'center',
         width: spinnerSize,
       }}>
-      <ErrorPanel
-        className={classes.errorPanel}
-        error="Something went wrong, please try again later."
-      />
       <CircularProgress />
     </div>
   );
--- a/src/components/criticalErrorMessage.jsx
+++ b/src/components/criticalErrorMessage.jsx
@@ -16,6 +16,7 @@ export const CriticalErrorMessage = () => (
     </a>
   </p>
 );
+
 export const DefaultErrorMessage = props => (
   <ErrorPanel
     className={props.style}
@@ -28,4 +29,10 @@ export const MissingDataErrorMessage = props => (
     className={props.style}
     error="This item has been missing data for at least 3 days."
   />
+);
+
+export const IvalidUrlMessage = () => (
+  <p style={{ textAlign: 'center', fontSize: '1.5em' }}>
+    <span>You have entered an invalid URL.</span>
+  </p>
 );
--- a/src/components/criticalErrorMessage.jsx
+++ b/src/components/criticalErrorMessage.jsx
@@ -1,8 +1,10 @@
 import React from 'react';
+import ErrorPanel from '@mozilla-frontend-infra/components/ErrorPanel';
 
 const newIssue =
   'https://github.com/mozilla/firefox-health-dashboard/issues/new';
-const CriticalErrorMessage = () => (
+
+export const CriticalErrorMessage = () => (
   <p style={{ textAlign: 'center', fontSize: '1.5em' }}>
     <span>
       There has been a critical error. We have reported it. If the issue is not
@@ -14,5 +16,16 @@ const CriticalErrorMessage = () => (
     </a>
   </p>
 );
+export const DefaultErrorMessage = props => (
+  <ErrorPanel
+    className={props.style}
+    error="Something went wrong, please try again later."
+  />
+);
 
-export default CriticalErrorMessage;
+export const MissingDataErrorMessage = props => (
+  <ErrorPanel
+    className={props.style}
+    error="This item has been missing data for at least 3 days."
+  />
+);
--- a/src/components/criticalErrorMessage.jsx
+++ b/src/components/criticalErrorMessage.jsx
@@ -31,7 +31,7 @@ export const MissingDataErrorMessage = props => (
   />
 );
 
-export const IvalidUrlMessage = () => (
+export const InvalidUrlMessage = () => (
   <p style={{ textAlign: 'center', fontSize: '1.5em' }}>
     <span>You have entered an invalid URL.</span>
   </p>
--- a/src/containers/PerfherderGraphContainer/index.jsx
+++ b/src/containers/PerfherderGraphContainer/index.jsx
@@ -33,8 +33,7 @@ class PerfherderGraphContainer extends Component {
     try {
       this.setState({ isLoading: true });
       this.setState(await getPerferherderData(series));
-      this.setState({ isLoading: false });
-    } catch (error) {
+    } finally {
       this.setState({ isLoading: false });
     }
   }
--- a/src/containers/PerfherderGraphContainer/index.jsx
+++ b/src/containers/PerfherderGraphContainer/index.jsx
@@ -22,6 +22,7 @@ const styles = () => ({
 class PerfherderGraphContainer extends Component {
   state = {
     data: null,
+    isLoading: false,
   };
 
   async componentDidMount() {
@@ -29,12 +30,18 @@ class PerfherderGraphContainer extends Component {
   }
 
   async fetchSetData({ series }) {
-    this.setState(await getPerferherderData(series));
+    try {
+      this.setState({ isLoading: true });
+      this.setState(await getPerferherderData(series));
+      this.setState({ isLoading: false });
+    } catch (error) {
+      this.setState({ isLoading: false });
+    }
   }
 
   render() {
     const { classes, title } = this.props;
-    const { data, jointUrl, options } = this.state;
+    const { data, jointUrl, options, isLoading } = this.state;
 
     return (
       <div key={title}>
@@ -49,6 +56,7 @@ class PerfherderGraphContainer extends Component {
         <ChartJsWrapper
           type="line"
           data={data}
+          isLoading={isLoading}
           options={options}
           missingDataError
         />
--- a/src/containers/RedashContainer/index.jsx
+++ b/src/containers/RedashContainer/index.jsx
@@ -18,6 +18,7 @@ const styles = {
 class RedashContainer extends Component {
   state = {
     datasets: null,
+    isLoading: false,
   };
 
   static propTypes = {
@@ -59,16 +60,22 @@ class RedashContainer extends Component {
   }
 
   async fetchSetState({ dataKeyIdentifier, redashDataUrl }) {
-    const redashData = await fetchJson(redashDataUrl);
+    try {
+      this.setState({ isLoading: true });
+      const redashData = await fetchJson(redashDataUrl);
 
-    this.setState({
-      datasets: telemetryDataToDatasets(redashData, dataKeyIdentifier),
-    });
+      this.setState({
+        datasets: telemetryDataToDatasets(redashData, dataKeyIdentifier),
+      });
+      this.setState({ isLoading: false });
+    } catch (error) {
+      this.setState({ isLoading: false });
+    }
   }
 
   render() {
     const { classes, options, redashQueryUrl, title } = this.props;
-    const { datasets } = this.state;
+    const { datasets, isLoading } = this.state;
 
     return (
       <div>
@@ -76,6 +83,7 @@ class RedashContainer extends Component {
           title={title}
           type="line"
           data={datasets}
+          isLoading={isLoading}
           options={options}
         />
         <div className={classes.linkContainer}>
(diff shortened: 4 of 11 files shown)


src/components/ChartJsWrapper/index.jsx
```diff
@@ -73,6 +73,10 @@ const ChartJsWrapper = ({
         textAlign: 'center',
         width: spinnerSize,
       }}>
+      <ErrorPanel
+        className={classes.errorPanel}
+        error="Something went wrong, please try again later."
```
sarah-clements: You can store this message in a variable such as `const defaultErrorMessage` and put it into a constants or helpers file since you are using it more than once.

src/telemetry/graph.jsx
```diff
@@ -32,6 +45,7 @@ export default class TelemetryContainer extends React.Component {
       this.graphSubtitleEl.textContent = graphData.description;
       this.graphEvolutionsTimeline(graphData, this.graphEl);
     } catch (error) {
+      this.setState({ errorFlag: true });
```
sarah-clements: I would rename `errorFlag` to something that more clearly denotes it's a boolean, such as `showErrorMessage`.

klahnakoski: I suggest renaming `errorFlag` to plain `error` so any value can be assigned, specifically a more detailed description of what happened.  `error != null` is an error.

aimenbatool: @klahnakoski got it. I will do it in the next push. PR is ready for review with requested changes.

cc: @sarah-clements

sarah-clements: > I suggest renaming `errorFlag` to plain `error` so any value can be assigned, specifically a more detailed description of what happened. `error != null` is an error.

True, I didn't suggest this previously since it seemed only 5xx errors would be relevant here and then it's a matter of messaging. So a generic "Something went wrong, please try again later." vs "NetworkError when attempting to fetch resource".

src/views/QuantumTP6/index.jsx
```diff
@@ -31,36 +32,39 @@ class TP6 extends React.Component {
...
-                      .map(s => ({ label: s.label, seriesConfig: s }))
-                      .toArray()}
-                  />
-                </Grid>
-              ))}
-          </Grid>
-        </DashboardPage>
-      </div>
-    );
+    if (bits === '32' || bits === '64')
```
klahnakoski: I am wondering out loud:

It seems messy to pollute the component with these checks.  Not just here, but it is happening in other parts of this PR too,  Can we let the Component throw an error, and let an HOC show the `InvalidUrlMessage`?  We may need to distinguish between user error (like in this case with bad URL)  and network error (which happens often to me).

We must still do a verification, but it need only be one line:

    if (bits === '32' || bits === '64') throw new Error("bad bits!");

I am currently attempting an HOC here https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/313/files#diff-21b4dce6876c43d2d41fa75031291026R11

aimenbatool: @klahnakoski I think I got what you're trying to say. But for my clarification, you mean that we should make a HOC and use that `<InvalidUrlMessage />` component there based on the message returned by wrapped component. 

>We may need to distinguish between user error (like in this case with bad URL) and network error (which happens often to me).

I didn't get it much. May be we need to use HOC only where there is user error ? 

Looking into your HOC, looks like time has come to dig into HOC now :-)

sarah-clements: > > We may need to distinguish between user error (like in this case with bad URL) and network error (which happens often to me).
> 
> I didn't get it much. May be we need to use HOC only where there is user error ?

@aimenbatool Creating a HOC is one possible solution in general - distinguishing between types of errors isn't specific to using a HOC. I suggest you read up on the advantages and potential pitfalls of using HOCs (not just understanding how they work) because it's important to understand the trade-offs of using one approach over another. :)

There are a few approaches you can take to handle user errors (http 400) - whether the user entered an invalid value or query parameter:
1) verify a value/parameter is valid _before_ making an API request and show a useful message to the user (for example, you can check for valid date formats or character length or whether a value is a number). With react router you can access parameters with the `location` prop.
2) You won't be able to verify every query parameter before it is used to make a request, and so the API may return a http 400 - bad request or http 404 - not found. Depending on how the API is configured, it should return a (hopefully) useful json message that can be shown to the user. Here's [an example ](https://treeherder.mozilla.org/api/performance/summary/ ) with `performance/summary` API.

The Fetch API has a few limitations regarding error handling, in that it will only reject a Promise for network related errors (5**) - it won't reject 4** errors. One way around this is to create a wrapper around `fetch` and check for the response.ok. If there is a 4** error that is likely returning a json response for the error, you can process that accordingly. For 5** errors, whether a json or text/html response is returned might be based on how the server is configured. You can find several articles on this topic (http status codes and how/what to show to the user) on the internet. It's worth reading up on this topic generally since there are different ways to approach it.

klahnakoski: @sarah-clements  please, feel free to contradict me; At the very least, there is information in our disagreement that I can benefit from. 

You bring up this is a routing issue: Maybe the error handling belongs in `<BrowserRouter>` as some more specific URL rules.  Maybe the error handling belongs in `withNavigation()`: which is responsible for URL parameters. Maybe it belongs in an HOC: that attempts to do its best to provide a humane feedback when failure happens.  Maybe error handling belongs on all three, or belongs in none.  

Your points about HOCs getting tricky may be valid.  HOCs should not be changing behaviour of the child components so I do not think this is true. However, I have not used HOCs enough to say if that is an unrealistic idealism, so you correct to point out there may be dragons there.

On the very specific subject of long HOC lines, like your example:

    withUserError(withLoading(withRouter(withStyles(styles)(Output))))

I do not think this long line is a problem; it think it is just aside effect of functional notation being a programmer's default, and not much consideration given to long chains of HOCs.  I would expect we eventually move to chaining the HOCs for readability.

```js
with(Output)
  .withStyles(styles)
  .withNavigation(config)
  .withRouter()
  .withUserError() 
;
```


At the very least, I would like to *attempt* removing this pattern from our components:

```js
error ? <ErrorComponent> : <GoodComponent>
```

That pattern has various forms, but it seems to pollute the code because it is everywhere.  I can be wrong about this because it is frontend code.  For backend code, when I see an `if` statement, it often indicates a bad data model. Many times these models are out of your control, so `if` statements are required; otherwise, we should be inspecting **why** the data is irregular and if we can fix it; if we can not fix it, because it is fundamental to the domain, then we should encapsulate the irregularity in a more regular data structure. 

For example, I am looking at the perfherder code so we can pull more data with less requests:  Here is each if statment and how to deal with it.

---
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/951b50ff431fbc3f2fe181437931b0940c7941ea/src/utils/perfherder/subbenchmarks.jsx#L27

This is filtering results, it is better to show it as such:

    toPairs(signatures).where({suite})....

---

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/951b50ff431fbc3f2fe181437931b0940c7941ea/src/utils/perfherder/subbenchmarks.jsx#L43

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/951b50ff431fbc3f2fe181437931b0940c7941ea/src/utils/perfherder/subbenchmarks.jsx#L45

These two are also filtering, but the hard way: matching hashes, then filtering the options found. The data structure should have joined the two tables on the hash, and then added to the filter above:

    toPairs(signatures).where({suite, option})....

---

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/951b50ff431fbc3f2fe181437931b0940c7941ea/src/utils/perfherder/subbenchmarks.jsx#L53

I consider this a legitimate check because it is an `assert`: We should be able to remove this code, but we do not for safety.

---

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/951b50ff431fbc3f2fe181437931b0940c7941ea/src/utils/perfherder/subbenchmarks.jsx#L87

Assume `percentileThreshold <= 100` and remove this if statement.

---

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/951b50ff431fbc3f2fe181437931b0940c7941ea/src/utils/perfherder/subbenchmarks.jsx#L111

The data coming out of Perfherder has an irregularity: The aggregate of the subtests is treated like a subtest. So we have if statements that checks for this, and you can see them littered throughout the code.  Even so, this `if` should not exist: any caller must deal with the aggregate separate from the subtest, and will have a symbiotic `if` statement on it's end; making *two* if statements to handle the same data irregularity.  I am guessing this `if` exists so one perfherder request is made instead of two. In this case, I suggest changing the code so the caller decides the signatures needed, and let this function simply retrieve them blindly.

---

klahnakoski: BTW, I do not intend to block this PR. If it works, then it is good enough for me!  The good thing about working code, beyond the fact it *works*; it represents use cases for later refactoring efforts.  We can attempt other error-handling strategies later.

sarah-clements: > You bring up this is a routing issue: Maybe the error handling belongs in `<BrowserRouter>` as some more specific URL rules. Maybe the error handling belongs in `withNavigation()`: which is responsible for URL parameters. Maybe it belongs in an HOC: that attempts to do its best to provide a humane feedback when failure happens. Maybe error handling belongs on all three, or belongs in none.

I think this would be a good use case for having an alternate view in `<BrowserRouter>` that shows 'Not found - did you mean x or x?' (or some other implementation). 

> I do think your points about HOCs getting tricky may be valid. HOCs should not be changing behaviour of the child components so I do not think this is true. However, I have not used HOCs enough to say if that is an unrealistic idealism, so you correct to point out there may be dragons there.

I'm not opposed to them generally - I have used it myself where there was stateful logic that could be shared with two different views, but I didn't want to hardwire the layout because there were some differences and I wanted to maintain flexibility.  The apprehension was regarding using multiple HOC for different aspects of error handling that could have the potential for prop name collisions and might be more efficient to combine.

> That pattern has various forms, but it seems to pollute the code because it is everywhere. I can be wrong about this because it is frontend code. For backend code, when I see an `if` statement, it often indicates a bad data model. Many times these models are out of your control, so `if` statements are required; otherwise, we should be inspecting **why** the data is irregular and if we can fix it; if we can not fix it, because it is fundamental to the domain, then we should encapsulate the irregularity in a more regular data structure.
> 
Yes, since this particular logic is reproduced everywhere so it does make sense to encapsulate it if it can be done simply. I've had these thoughts myself regarding error handling in Treeherder. I see a challenge in how much of the different error handling to unite and which to keep separate (with ErrorBoundary components, query param validation, API request failures, etc!)

I appreciate having these discussions and I think it's also useful for @aimenbatool to be able to see the 'why' behind a decision. I'm a fan of starting simply and refactoring as needed, which was why I was trying to encourage that direction for her. I think it'd be a great learning opportunity to land this working code as you've suggested above, and for her to work with you on different error handling strategies :)

aimenbatool: Great discussion @sarah-clements and @klahnakoski! 
Although it took me time to absorb and try to understand, it is good to see how teams discuss things and make decisions. Somethings are still alien and overwhelming to me in the discussion as a newbie. 
I would like to pick the suggestion 'land this code and refactor step by step'. It will help me not to overwhelm and get confused with a lot of things coming at the same time. :-)  I would love to work on error handling strategies but  first I need to get my self more comfortable playing with HOCs on simple things.

klahnakoski:
@aimenbatool if this is ready for review and merge, please remove the "WIP"

aimenbatool:
@klahnakoski it is ready for review and merge.

aimenbatool:
@klahnakoski, @sarah-clements it is ready for review and merge.

src/components/criticalErrorMessage.jsx
```diff
@@ -15,4 +17,22 @@ const CriticalErrorMessage = () => (
...
+);
+
+export const MissingDataErrorMessage = props => (
+  <ErrorPanel
+    className={props.style}
+    error="This item has been missing data for at least 3 days."
+  />
+);
+
+export const IvalidUrlMessage = () => (
```
sarah-clements: This should be spelled `InvalidUrlMessage` :)

src/components/criticalErrorMessage.jsx
```diff
@@ -15,4 +17,22 @@ const CriticalErrorMessage = () => (
...
+  <ErrorPanel
+    className={props.style}
+    error="Something went wrong, please try again later."
+  />
+);
+
+export const MissingDataErrorMessage = props => (
+  <ErrorPanel
+    className={props.style}
+    error="This item has been missing data for at least 3 days."
```
sarah-clements: Instead of creating two components that are doing the same thing, I would store these two error messages in variables or as keys in an object and pass those in as props directly to `ErrorPanel`, or have some conditional logic that determines which message to show based on the error code.

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -22,19 +22,26 @@ const styles = () => ({
...
   }
 
   async fetchSetData({ series }) {
-    this.setState(await getPerferherderData(series));
+    try {
+      this.setState({ isLoading: true });
+      this.setState(await getPerferherderData(series));
+      this.setState({ isLoading: false });
+    } catch (error) {
+      this.setState({ isLoading: false });
```
sarah-clements: Shouldn't we do something with this error - store it in state and then use it to show `ErrorPanel`? 

`this.setState({ isLoading: false });` can be moved outside the try/catch block since it will be called either way.

src/containers/RedashContainer/index.jsx
```diff
@@ -59,23 +60,30 @@ class RedashContainer extends Component {
...
+      this.setState({ isLoading: true });
+      const redashData = await fetchJson(redashDataUrl);
 
-    this.setState({
-      datasets: telemetryDataToDatasets(redashData, dataKeyIdentifier),
-    });
+      this.setState({
+        datasets: telemetryDataToDatasets(redashData, dataKeyIdentifier),
+      });
+      this.setState({ isLoading: false });
```
sarah-clements: Same comment as above - combine state updates as much as possible.

klahnakoski: Same as before, but the `isLoading: false` into the finally clause

src/containers/PerfherderGraphContainer/index.jsx
sarah-clements: Technically it may behave asynchronously, but that's not why I was mentioning it :) Calling setState out of an event handle used to cause a new render for each time it's called but that is apparently not the case as of v16+: https://itnext.io/react-setstate-usage-and-gotchas-ac10b4e03d60

aimenbatool:
> this looks fine to me. Some local optimizations (like using the `finally` clause), but they can wait until later.
> 
> My big concern is setting the `isLoading`, but we do not use them. What is the plan?

We are using isLoading [here](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/309/files#diff-a6d9507721ad2f63c84d8f75de14ff24R34).

src/containers/PerfherderGraphContainer/index.jsx
aimenbatool: I have pushed the updated code. Please let me know if it looks good now.

klahnakoski:
@aimenbatool looks good, please merge.

src/containers/PerfherderGraphContainer/index.jsx
sarah-clements: The changes look good Aimen, but I'm still wondering why we aren't catching the errors and showing an ErrorPanel here and in the BugzillaGraph component?

klahnakoski: @sarah-clements the error catching will be caught in my next PR: https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/330

Issue #288, opened 2019-02-04
Discussion: The nimblegraph complexity


klahnakoski:
Navigating the nimblegraph components is a nightmare. My IDE lets me jump around the definitions fine, but the browser debuggers are primitive; forcing me to recall the locations of each of the React components so I can set break points. Stepping into code does not work because react has inserted multiple functions between every nimbledroid function.

Here is a diagram that demonstrated the problem:

![image](https://user-images.githubusercontent.com/2334429/52220934-06605180-286e-11e9-92f7-7af43d6a1159.png)

Arrows show dependencies, and only the dependencies exclusive to the nimbledroid code (libraries and config needed by other components are not included for simplicity).

There are some problems I see:
* A nimbledroid components are spread over multiple directories, making it near-impossible to reuse in another dashboard.
* The `components` are used by just one `container`; they should be in-lined, or at least kept in the container file.
* `BackendClient` is some attempt at isolation, but it is only used by nimbledroid so it only serves to disperse the data fetching logic
* `utils/nimbledroid/index.jsx` is not a component, rather some configuration plus some data transformation logic. It should be merged with `fetchNimbledroidData`
* There are multiple directories with a single `index.jsx`:  I do not know why this pattern exists, but it makes browser debug navigation just a bit more painful.

These problems are repeated over the whole of the project, and I would like to clean it up, over time, as we touch each piece of code (if ever).  I would like to use the following principal 

> Only use indirection if more than one component/method/object must reference it.

This means we should not be making `index.jsx` files unless we need another file in the same directory.  We should not make component files when only one component file needs the component, we should consider in lining components if the component is used in only one place. Methods should be in-lined if used by only one method, files should be combined if used by only one other file. 

But first, @eliperelman, @sarah-clements, @aimenbatool, I would like all your input.

sarah-clements:
>  We should not make component files when only one component file needs the component, we should consider in lining components if the component is used in only one place.

I'll admit I'm not as familiar with the code base but I think potential problems with adopting this approach is that files could become unwieldy if all of a parent component's child components were in the same file. It would also make it difficult to see what components can be reused elsewhere. In the Treeherder codebase, we generally put one component to a file and generic, shared components go in a ./shared directory (as do helper functions, depending on if they're used elsewhere in the app or only by one particular app - TH, Perfherder, Intermittent Failures View).

Putting multiple pure (stateless) components per file might be a good compromise if they are grouped together logically and the file name is clear.

[...]

PR #313, opened 2019-02-20
Navigation for charts

klahnakoski:


klahnakoski:
@helfi92 

The important part is the addition of two a JSON config that drives the selectors:

for tp6: https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/313/files?utf8=%E2%9C%93&diff=split&w=1#diff-a39322f505fbfbbfe383aa2d2aa63a56R73
for tp6m: https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/313/files?utf8=%E2%9C%93&diff=split&w=1#diff-acdeeeec047caee18cba7cd1df8861acR71

here is the `withNavigator()` method, which is an HOC:

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/313/files?utf8=%E2%9C%93&diff=split&w=1#diff-21b4dce6876c43d2d41fa75031291026R11

image

klahnakoski:
@djmitche  Since this is my first higher-order-component, I like your feedback on what I have done.  Your review is not required, just appreciated.  Thank you!

package.json
```diff
@@ -52,6 +53,7 @@
     "react-chartjs-2": "^2.7.4",
     "react-dom": "^16.5.0",
     "react-hot-loader": "^4.3.6",
+    "react-loadable": "^5.5.0",
```
helfi92: Where is this being used? Also, for lazy loading, you may want to look into `React.lazy` instead.

src/vendor/components/navigation/index.jsx
```diff
@@ -0,0 +1,132 @@
+// eslint-disable react/prop-types
```
helfi92: I would typically avoid putting a HOC component under a `components` directory since it's not really a UI component. Perhaps a `utils` folder.

src/vendor/components/navigation/index.jsx
```diff
@@ -0,0 +1,132 @@
...
+        return (
+          <div className={classes.root}>
+            {config.map(c => {
+              const { type, id, label, options } = c;
+
+              return React.createElement(type, {
+                key: id,
+                id,
+                label,
+                handleChange: (...args) => self.onPathChange(...args),
```
helfi92: You should be able to use `this` instead of `self` by changing `navComponents() {` to be `navComponents = () => {`.

src/vendor/components/navigation/index.jsx
```diff
@@ -0,0 +1,132 @@
...
+          />
+        );
+      }
+    }
+
+    Output.propTypes = {
+      classes: PropTypes.shape().isRequired,
+      config: PropTypes.arrayOf(
+        PropTypes.shape({
+          type: PropTypes.instanceOf(React.Component),
```
helfi92: For prop types that are not required, it's good practice to define defaults for them in the static method `defaultProps`.

src/vendor/components/navigation/index.jsx
```diff
@@ -0,0 +1,132 @@
...
+          ),
+        })
+      ),
+    };
+
+    const styles = () => ({
+      root: {
+        display: 'flex',
+        textAlign: 'left',
+        padding: '15px',
```
helfi92: `padding: 15,`

src/vendor/components/navigation/slider.jsx
```diff
@@ -0,0 +1,96 @@
...
+import Tooltip from '@material-ui/core/Tooltip';
+
+const styles = () => ({
+  root: {
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
+    maxWidth: 305,
+    margin: '0 20px',
+    padding: '15px',
```
helfi92: `padding: 15,`

src/vendor/components/navigation/slider.jsx
```diff
@@ -0,0 +1,96 @@
...
+    textAlign: 'left',
+  },
+  label: {
+    whiteSpace: 'nowrap',
+    padding: '0 10px 3px 0',
+    fontSize: '12px',
```
helfi92: `fontSize: 12,`

src/vendor/components/navigation/slider.jsx
```diff
@@ -0,0 +1,96 @@
+import React, { Component } from 'react';
```
helfi92: Rename the file so that it matches the casing of the class: `src/vendor/components/navigation/Slider.jsx`.

src/vendor/components/navigation/picker.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
```
helfi92: Rename the file to `Picker.jsx` to match the class name.

src/views/QuantumTP6m/index.jsx
```diff
@@ -19,27 +21,16 @@ const styles = {
...
-    const { test, platform } = this.state;
+    const { classes, navigation, test, platform } = this.props;
     const subtitle = frum(TP6_TESTS)
       .where({ id: test })
       .first().label;
 
     return (
       <div className={classes.body}>
         <DashboardPage key={subtitle} title="TP6 Mobile" subtitle={subtitle}>
+          {navigation}
```
helfi92: If `navigation` is included in every page, then perhaps it would make sense to add this line in `DashboardPage`.

src/vendor/components/navigation/index.jsx
```diff
@@ -0,0 +1,132 @@
...
+  //
+  // Expects a `config` that is a list of configurations for various
+  // selection widgets. Right now there is only Picker.
+  //
+  // Adds properties to `props`:
+  // * `navigation` - that contains a Navigation component
+  //   to include on `render()`
+  // * properties with names frum(config).select("id")
+
+  return WrappedComponent => {
```
helfi92: I think we'll want to return the class otherwise we would be returning nothing:

```jsx
// no braces at the end
return WrappedComponent =>
  class Output extends React.Component {
```

src/vendor/components/navigation/index.jsx
klahnakoski: I have no idea the difference between "component", "container", "view", or "hoc" types; they are all "components" to me.  As long as the directory structure is a rough estimate of component dependencies I am happy.  Maybe the name is wrong; I am open to suggestions.

I think the [nimbledroid mess](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/288) was caused by separating component types into directories rather than keeping one component, and its children, together.

src/views/QuantumTP6m/index.jsx
klahnakoski: Still too early to say; these particular pages are a uniform series of charts, while the others are heterogeneous.

src/vendor/components/navigation/index.jsx
klahnakoski: We `return withRouter(withStyles(styles)(Output));`  which is need to provide `history` and styling to the navigation component.

src/vendor/components/navigation/index.jsx
klahnakoski: I was attempting `handleChange: this.onPathChange`, but function is not bound to `this` at call time.

I changed 

    onPathChange(event) {

to

    onPathChange = event => {

which seemed to solve the binding problem

src/vendor/components/navigation/index.jsx
klahnakoski: anyway, you already suggested `utils` which is good enough for me.

klahnakoski:
@helfi92  Thank you for another good review: I am still learning.  Anything else in this PR?

PR #317, opened 2019-02-23
reduce size

klahnakoski:
by adding in the mozilla-frontend-infra/compnenets to vendor, and tweaking the dependencies, we loose about a megabyte


src/components/ChartJsWrapper/index.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import Chart from 'react-chartjs-2';
 import CircularProgress from '@material-ui/core/CircularProgress';
 import { withStyles } from '@material-ui/core/styles';
-import ErrorPanel from '@mozilla-frontend-infra/components/ErrorPanel';
+import CriticalErrorMessage from '../criticalErrorMessage';
```
klahnakoski: this is the only important change, all the rest is pulling in the components so I could track the dependencies.

PR #318, opened 2019-02-24
WIP: Create custom tooltips

### LINKED ISSUE #303: Add hg and job link inside tooltip metadata
Please refer to this https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/213#issuecomment-462912268 for more details.

2alin:
Fixes  #303 

@klahnakoski, I'm not sure if I'm doing it right, but here is what I have done so far:

I've defined a function `generateCustomTooltip` that creates a container, place it in the right position, injects it into the DOM and imports a `CustomTooltip` component which received all the information needed in `tooltipModel` to fill the tooltip with data.

1. First of all, I'm still wondering if using react component for the tooltip is a right choice, mainly because the way the tooltip appears now, when I move the pointer through the data points, is all laggy, choppy [1]. In contrast with how it was when I used only HTML injection and no react components [2] [3]. Maybe it was the way how I implemented it. But this is a **crucial issue**.

2. Also, I'm using `dangerouslySetInnerHTML` to inject the needed HTML for the footer's links. I found that I could use [html-react-parser](https://www.npmjs.com/package/html-react-parser), which one should I use?

3. Which key should I use for every line in the title/body/footer, because I'm mapping them inside the component.

4. I'm aware that either the directory where I located the files, the extensions, or the way I created them is not in sync with the rest of the project. So, any help there would be highly appreciated.  

Thank you :)

---

[1] very frustrating to make the tooltip appears, with the current implementation
ezgif-4-a65c7df5b081

[2] previous code I used to generate custom tooltip (no react)
https://gist.github.com/2alin/12abc8dd1ae9c92d0e7d82f03899b6b1

[3] it was smooth when I used the previous approach ([1])
ezgif com-video-to-gif

### DIFF
--- a/src/utils/chartJs/CustomTooltip.jsx
+++ b/src/utils/chartJs/CustomTooltip.jsx
@@ -14,12 +14,6 @@ const styles = {
     height: '10px',
     marginRight: '10px',
   },
-  table: {
-    borderCollapse: 'collapse',
-    '& p': {
-      margin: 0,
-    },
-  },
 };
 
 function Footer({ dataset, index, currData }) {
@@ -58,17 +52,15 @@ function CustomTooltip({ classes, tooltipModel, series }) {
 
   return (
     <div className={classes.tooltip} style={paddingStyle}>
-      <div className={classes.table}>
-        <div>{xLabel}</div>
-        <div>
-          <span style={labelColors} className={classes.tooltipKey} />
-          {seriesLabel}: {currData}
-        </div>
-        <div>
-          {currData} ({higherOrLower})
-        </div>
-        {index > 0 && <Footer {...{ dataset, index, currData }} />}
+      <div>{xLabel}</div>
+      <div>
+        <span style={labelColors} className={classes.tooltipKey} />
+        {seriesLabel}: {currData}
       </div>
+      <div>
+        {currData} ({higherOrLower})
+      </div>
+      {index > 0 && <Footer {...{ dataset, index, currData }} />}
     </div>
   );
 }
--- a/src/utils/chartJs/CustomTooltip.jsx
+++ b/src/utils/chartJs/CustomTooltip.jsx
@@ -3,18 +3,20 @@ import { withStyles } from '@material-ui/core/styles';
 import Fade from '@material-ui/core/Grow';
 
 const styles = {
-  table: {
+  tooltip: {
     background: 'rgba(0, 0, 0, .8)',
     color: 'white',
     borderRadius: '4px',
     pointerEvents: 'none',
-    borderCollapse: 'collapse',
   },
   tooltipKey: {
     display: 'inline-block',
     width: '10px',
     height: '10px',
     marginRight: '10px',
+  },
+  table: {
+    borderCollapse: 'collapse',
   },
 };
 
@@ -27,44 +29,51 @@ function CustomTooltip({ classes, tooltipModel }) {
 
   return (
     <Fade in={tooltipModel.opacity === 1}>
-      <table className={classes.table}>
-        <thead>
-          {titleLines.map(title => (
-            <tr>
-              <th>{title}</th>
-            </tr>
-          ))}
-        </thead>
-        <tbody>
-          {bodyLines.map((body, i) => {
-            const { backgroundColor, borderColor } = tooltipModel.labelColors[
-              i
-            ];
-
-            return (
+      <div
+        className={classes.tooltip}
+        style={{
+          padding: `${tooltipModel.yPadding}px ${tooltipModel.xPadding}px`,
+        }}
+      >
+        <table className={classes.table}>
+          <thead>
+            {titleLines.map(title => (
               <tr>
-                <td>
-                  <span
-                    style={{
-                      background: backgroundColor,
-                      borderColor,
-                    }}
-                    className={classes.tooltipKey}
-                  />
-                  {body}
-                </td>
+                <th>{title}</th>
               </tr>
-            );
-          })}
-        </tbody>
-        <tfoot>
-          {footerLines.map(footer => (
-            <tr>
-              <td dangerouslySetInnerHTML={{ __html: footer }} />
-            </tr>
-          ))}
-        </tfoot>
-      </table>
+            ))}
+          </thead>
+          <tbody>
+            {bodyLines.map((body, i) => {
+              const { backgroundColor, borderColor } = tooltipModel.labelColors[
+                i
+              ];
+
+              return (
+                <tr>
+                  <td>
+                    <span
+                      style={{
+                        background: backgroundColor,
+                        borderColor,
+                      }}
+                      className={classes.tooltipKey}
+                    />
+                    {body}
+                  </td>
+                </tr>
+              );
+            })}
+          </tbody>
+          <tfoot>
+            {footerLines.map(footer => (
+              <tr>
+                <td dangerouslySetInnerHTML={{ __html: footer }} />
+              </tr>
+            ))}
+          </tfoot>
+        </table>
+      </div>
     </Fade>
   );
 }
--- a/src/utils/chartJs/CustomTooltip.jsx
+++ b/src/utils/chartJs/CustomTooltip.jsx
@@ -21,26 +21,29 @@ const styles = {
     },
   },
 };
-const Header = ({ xLabel }) => <p>{xLabel}</p>;
-const TopBodyLines = props => (
-  <React.Fragment>
-    <p>
-      <span style={props.labelColors} className={props.classes.tooltipKey} />
-      {props.seriesLabel}: {props.currData}
-    </p>
-    <p>
-      {props.currData} ({props.higherOrLower})
-    </p>
-  </React.Fragment>
-);
+
+function Footer({ dataset, index, currData }) {
+  const prevData = dataset[index - 1].value;
+  const prevRevision = dataset[index - 1].revision;
+  const delta = (currData - prevData).toFixed(2);
+  const deltaPercentage = ((currData / prevData - 1) * 100).toFixed(2);
+  const currRevision = dataset[index].revision;
+  const hgURL = `https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=${prevRevision}&tochange=${currRevision}`;
+
+  return (
+    <React.Fragment>
+      <div>
+        Δ {delta} ({deltaPercentage} %)
+      </div>
+      <div>
+        <a href={hgURL}>{currRevision.slice(0, 12)}</a>
+      </div>
+    </React.Fragment>
+  );
+}
 
 function CustomTooltip({ classes, tooltipModel, series }) {
-  if (tooltipModel.opacity === 0) {
-    // Chartjs removes some properties in `tooltipModel` when datapoint
-    // is not active. Those propertiles includes `title`, `dataPoints`, etc.
-    // We can't go forward if that's the case
-    return null;
-  }
+  if (tooltipModel.opacity === 0) return null;
 
   const { xLabel, index, datasetIndex } = tooltipModel.dataPoints[0];
   const dataset = series[datasetIndex].data;
@@ -52,48 +55,19 @@ function CustomTooltip({ classes, tooltipModel, series }) {
   const paddingStyle = {
     padding: `${tooltipModel.yPadding}px ${tooltipModel.xPadding}px`,
   };
-  const cmpProps = {
-    classes,
-    labelColors,
-    seriesLabel,
-    currData,
-    higherOrLower,
-  };
-
-  // console.log('series', series);
-  // console.log('tooltipModel', tooltipModel);
-
-  if (index === 0) {
-    return (
-      <div className={classes.tooltip} style={paddingStyle}>
-        <div className={classes.table}>
-          <Header xLabel={xLabel} />
-          <TopBodyLines {...cmpProps} />
-        </div>
-      </div>
-    );
-  }
-
-  const prevData = dataset[index - 1].value;
-  const delta = (currData - prevData).toFixed(2);
-  const deltaPercentage = ((currData / prevData - 1) * 100).toFixed(2);
-  const currRevision = dataset[index].revision;
-  const prevRevision = dataset[index - 1].revision;
-  const hgURL = `https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=${prevRevision}&tochange=${currRevision}`;
 
   return (
     <div className={classes.tooltip} style={paddingStyle}>
       <div className={classes.table}>
-        <Header xLabel={xLabel} />
-        <TopBodyLines {...cmpProps} />
-
-        <p>
-          Δ {delta} ({deltaPercentage} %)
-        </p>
-
-        <p>
-          <a href={hgURL}>{currRevision.slice(0, 12)}</a>
-        </p>
+        <div>{xLabel}</div>
+        <div>
+          <span style={labelColors} className={classes.tooltipKey} />
+          {seriesLabel}: {currData}
+        </div>
+        <div>
+          {currData} ({higherOrLower})
+        </div>
+        {index > 0 && <Footer {...{ dataset, index, currData }} />}
       </div>
     </div>
   );
--- a/src/utils/chartJs/CustomTooltip.jsx
+++ b/src/utils/chartJs/CustomTooltip.jsx
@@ -28,27 +28,6 @@ function CustomTooltip({ classes, tooltipModel, series }) {
   const paddingStyle = {
     padding: `${tooltipModel.yPadding}px ${tooltipModel.xPadding}px`,
   };
-  let footer = null;
-
-  if (index > 0) {
-    const [prev, curr] = currSeries.data.slice(index - 1);
-    const delta = curr.value - prev.value;
-    const deltaPercentage = (delta / prev.value) * 100;
-    const hgURL = `https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=${
-      prev.revision
-    }&tochange=${curr.revision}`;
-
-    footer = (
-      <React.Fragment>
-        <div>
-          Δ {delta.toFixed(2)} ({deltaPercentage.toFixed(1)} %)
-        </div>
-        <div>
-          <a href={hgURL}>{curr.revision.slice(0, 12)}</a>
-        </div>
-      </React.Fragment>
-    );
-  }
 
   return (
     <div className={classes.tooltip} style={paddingStyle}>
@@ -60,7 +39,27 @@ function CustomTooltip({ classes, tooltipModel, series }) {
       <div>
         {currPoint.yLabel} ({higherOrLower})
       </div>
-      {footer}
+      {(() => {
+        if (index === 0) return null;
+
+        const [prev, curr] = currSeries.data.slice(index - 1);
+        const delta = curr.value - prev.value;
+        const deltaPercentage = (delta / prev.value) * 100;
+        const hgURL = `https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=${
+          prev.revision
+        }&tochange=${curr.revision}`;
+
+        return (
+          <React.Fragment>
+            <div>
+              Δ {delta.toFixed(2)} ({deltaPercentage.toFixed(1)} %)
+            </div>
+            <div>
+              <a href={hgURL}>{curr.revision.slice(0, 12)}</a>
+            </div>
+          </React.Fragment>
+        );
+      })()}
     </div>
   );
 }
--- a/src/utils/chartJs/CustomTooltip.jsx
+++ b/src/utils/chartJs/CustomTooltip.jsx
@@ -1,0 +1,72 @@
+import React from 'react';
+import { withStyles } from '@material-ui/core/styles';
+import Fade from '@material-ui/core/Grow';
+
+const styles = {
+  table: {
+    background: 'rgba(0, 0, 0, .8)',
+    color: 'white',
+    borderRadius: '4px',
+    pointerEvents: 'none',
+    borderCollapse: 'collapse',
+  },
+  tooltipKey: {
+    display: 'inline-block',
+    width: '10px',
+    height: '10px',
+    marginRight: '10px',
+  },
+};
+
+function CustomTooltip({ classes, tooltipModel }) {
+  const titleLines = tooltipModel.title || [];
+  const bodyLines = tooltipModel.body
+    ? tooltipModel.body.map(bodyItem => bodyItem.lines)
+    : [];
+  const footerLines = tooltipModel.footer || [];
+
+  return (
+    <Fade in={tooltipModel.opacity === 1}>
+      <table className={classes.table}>
+        <thead>
+          {titleLines.map(title => (
+            <tr>
+              <th>{title}</th>
+            </tr>
+          ))}
+        </thead>
+        <tbody>
+          {bodyLines.map((body, i) => {
+            const { backgroundColor, borderColor } = tooltipModel.labelColors[
+              i
+            ];
+
+            return (
+              <tr>
+                <td>
+                  <span
+                    style={{
+                      background: backgroundColor,
+                      borderColor,
+                    }}
+                    className={classes.tooltipKey}
+                  />
+                  {body}
+                </td>
+              </tr>
+            );
+          })}
+        </tbody>
+        <tfoot>
+          {footerLines.map(footer => (
+            <tr>
+              <td dangerouslySetInnerHTML={{ __html: footer }} />
+            </tr>
+          ))}
+        </tfoot>
+      </table>
+    </Fade>
+  );
+}
+
+export default withStyles(styles)(CustomTooltip);
--- a/src/utils/chartJs/CustomTooltip.jsx
+++ b/src/utils/chartJs/CustomTooltip.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { withStyles } from '@material-ui/core/styles';
-import Fade from '@material-ui/core/Grow';
 
 const styles = {
   tooltip: {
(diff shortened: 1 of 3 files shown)


src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -35,11 +37,30 @@ const generateInitialOptions = series => {
...
           return tooltipData;
         },
       },
+      enabled: false,
+      custom(tooltipModel) {
+        // eslint-disable-next-line no-underscore-dangle
+        const { canvas } = this._chart;
+
+        generateCustomTooltip(canvas, tooltipModel);
+      },
```
klahnakoski: these lines should read 

    custom: CustomTooltip

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -35,11 +37,30 @@ const generateInitialOptions = series => {
...
 
             tooltipData.push(`Δ ${delta} (${deltaPercentage}%)`);
+
+            const currRevision = series.data[tooltipItems[0].index].revision;
+            const prevRevision =
+              series.data[tooltipItems[0].index - 1].revision;
+            const hgLink = `https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=${prevRevision}&tochange=${currRevision}`;
+
+            tooltipData.push(
+              `<a href="${hgLink}" target="_blank">${currRevision.slice(
```
klahnakoski: instead of this format string, use JSX

              <a href={hgLink} target="_blank">{currRevision.slice(0,12)}</a>

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -35,11 +37,30 @@ const generateInitialOptions = series => {
             ).toFixed(2);
 
             tooltipData.push(`Δ ${delta} (${deltaPercentage}%)`);
+
```
klahnakoski: this whole `tooltips.callbacks.footer` method should go into the `CustomTooltip`

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <Fade in={tooltipModel.opacity === 1}>
+      <div
+        className={classes.tooltip}
+        style={{
+          padding: `${tooltipModel.yPadding}px ${tooltipModel.xPadding}px`,
+        }}
+      >
+        <table className={classes.table}>
+          <thead>
```
klahnakoski: do not render anything in the <table>`.  You will be able to find the `titleLines` (aka the data of the datapoint somewhere in the `tootipModel` structure.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,81 @@
...
+            {titleLines.map(title => (
+              <tr>
+                <th>{title}</th>
+              </tr>
+            ))}
+          </thead>
+          <tbody>
```
klahnakoski: probably always empty, do not render

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,81 @@
...
+                      }}
+                      className={classes.tooltipKey}
+                    />
+                    {body}
+                  </td>
+                </tr>
+              );
+            })}
+          </tbody>
+          <tfoot>
```
klahnakoski: instead of this footer, use `tooltips.callbacks.footer` code further down

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,81 @@
...
+            {footerLines.map(footer => (
+              <tr>
+                <td dangerouslySetInnerHTML={{ __html: footer }} />
+              </tr>
+            ))}
+          </tfoot>
+        </table>
```
klahnakoski: so, in summary, this table is removed, and replace with the footer function, plus a little more to show the date of the datapoint

src/utils/chartJs/CustomTooltip.jsx
2alin: @klahnakoski: Nope, it's never empty by default. I consoled log `tooltipModel`for you when a data point is hovered in [this gist](https://gist.github.com/2alin/2a1ef853d3dc9d251bb398b1f624c422). 

ChartJS (by default) uses `.title` for the x-values (in our case, the date), and builds `.body` using the dataset (series) label and the y-value (in our case, e.g. "Moto G5 (arm7): 257" ). So title and body are build by chartjs as usual, I'm only rendering them.

`.footer`was the one that was build by us (using something that chartjs calls 'fallback' in the docs) and because it contains anchors we are rendering them in a special way.

--------

Now we could build everything in one container (for example, everything in `.body`) but, we need to create empty callbacks for `.title`and `.footer` and create a long callback for `.body` that stores all the lines needed. And then we can render all the tooltip using only one map (bodyLines.map). What do you think? do you like this idea better?

klahnakoski: > ChartJS (by default) uses `.title` for the x-values (in our case, the date), and builds `.body` using the dataset (series) label and the y-value (in our case, e.g. "Moto G5 (arm7): 257" ). 

Ok.  Instead of rendering `body`, render the "dataset (series) label and the y-value" ourselves.  The idea is to avoid using the ChartJS entirely so no knowledge is required to understand the custom tooltip.  We may consider overriding the `body` function (`()=>null`) just so we do not run the default chartJS code 
 
> Now we could build everything in one container (for example, everything in `.body`) but, we need to create empty callbacks for `.title`and `.footer` and create a long callback for `.body` that stores all the lines needed. 

Yes, empty callbacks for `.title`and `.footer` are good. We will not need the `body` lines because we can get the info from the tooltipModel

    seriesName + " " +tooltipModel.dataPoints[0].yLabel

I have tried to find where `seriesName` can be pulled from (other than the `body`, but I can not seem to find it.

> And then we can render all the tooltip using only one map (bodyLines.map). What do you think? do you like this idea better?

I do not like this, but since I can not find where `seriesName` is in the chartJS datastructure, then I guess that's how it must be.

src/utils/chartJs/CustomTooltip.jsx
klahnakoski: This is a case where thinking about the future is not a good idea. Simply build the minimum.  I imagine future custom tooltips to be a copy-and-change of your custom tooltip.  You are doing the hard work of making the first one.

src/utils/perfherder/chartJs/perfherderFormatter.js
klahnakoski: Yes, leave it until later.  I see that the `tooltipModel` does not have all the information we require, we must attach some more before we call `CustomTooltip`

src/utils/chartJs/CustomTooltip.jsx
2alin: > The idea is to avoid using the ChartJS entirely so no knowledge is required to understand the custom tooltip.

Got it!!



> but since I can not find where `seriesName` is in the chartJS datastructure

This!! was the false alarm I was going to talk with you last night (along as many other values needed related to it. But it's  a false alarm, because we can use

2alin: > The idea is to avoid using the ChartJS entirely so no knowledge is required to understand the custom tooltip.

Got it!!



> but since I can not find where `seriesName` is in the chartJS datastructure

@klahnakoski This!! was the false alarm I was going to talk with you last night, as well as many other values needed related to it. But it's  a false alarm, because we can use  `series` from line 12: 
```const generateInitialOptions = series => {...} ```
it's passed all the way down to `tooltips.custom` :grin:

File: src/utils/perfherder/chartJs/perfherderFormatter.js (hunk 1 above)
Comment (2alin) [reply src/utils/perfherder/chartJs/perfherderFormatter.js]: No it hasn't, but there's a workaround, check my other comment please.

**Comment by 2alin** [comment]
@klahnakoski 
Main changes:
1. `perfherderFormatter.js` is way cleaner now, I moved all the `callback.footer`to `CustomTooltip` as suggested. Also, I was forced to modify some lines (37, 14, check review )a little

**Comment by 2alin** [comment]
@klahnakoski 
Main changes:
1. `perfherderFormatter.js` is way cleaner now, I moved all the `callback.footer`to `CustomTooltip` as suggested. Also, I was forced to modify some lines (37, 14, check review comment )a little, in order to pass all the series data to `CustomTooltip`.
2. `CustomTooltip.jsx` doesn't have mappings now, and renders everything accordingly. But! I got into a dilemma while trying to consider the case of the **first datapoint**, remember? we shouldn't render the delta lines for that point. So, my first try was to use '&&' operator while rendering like: 

```index > 0 && (jsx for deltaline, etc)```

unfortunately if we do that we need to initialize and later on assign the values for `prevData, delta, etx`, like this:

```
let prevData = null;
let delta = null;
let deltaPercentage = null;
(+3 lines more)
if (index > 0) {
  prevData = ...
  delta = ...;
  deltaPercentage = ...;
  (+3 linex more)
}
```

which is quite repetetive. We can't do this:

```
if (index > 0) {
  const prevData = ...
  const delta = ...;
  const deltaPercentage = ...;
  (+3 linex more)
}
```
because JSX needs the variables in the && operator to be all in the same scope as the operator.

So, I decided to not go that way, and then use conditionals to render two components in different ways (one for initial data point, and other for the rest), but it has its cons too. I ended up defining two minicomponents (Header, ToBodyLines) in order be less repetitive. 

Was my decision right? or should I go into another direction?
Thank you!

**Comment by 2alin** [comment]
Complementing my previous comment. 
There's also pending `generateCustomTooltop`to modify/refactor, I think that was one of your previous suggestion as well as the declaration inside `tooltips.custom`. We can work with that once you review the last changes I made. 
Thank you again :)

File: src/utils/chartJs/CustomTooltip.jsx [hunk 8]
```diff
@@ -0,0 +1,115 @@
...
+    display: 'inline-block',
+    width: '10px',
+    height: '10px',
+    marginRight: '10px',
+  },
+  table: {
+    borderCollapse: 'collapse',
+  },
+};
+const Header = ({ xLabel }) => (
```
klahnakoski: inline

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,115 @@
...
+  <thead>
+    <tr>
+      <th>{xLabel}</th>
+    </tr>
+  </thead>
+);
+const TopBodyLines = props => (
```
klahnakoski: inline

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,115 @@
...
+    higherOrLower,
+  };
+
+  // console.log('series', series);
+  // console.log('tooltipModel', tooltipModel);
+
+  if (index === 0) {
+    return (
+      <div className={classes.tooltip} style={paddingStyle}>
+        <table className={classes.table}>
```
klahnakoski: Remove `<table>`, `<td>`, `<tr>`, etc.  replace with a single `<div>` and add styling if required.  This will make my inline suggestion easier.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,115 @@
...
+    labelColors,
+    seriesLabel,
+    currData,
+    higherOrLower,
+  };
+
+  // console.log('series', series);
+  // console.log('tooltipModel', tooltipModel);
+
+  if (index === 0) {
```
klahnakoski: please inline this check, and the one line (not) shown:

    <div>
        <span style={props.labelColors} className={props.classes.tooltipKey} />
        {props.seriesLabel}: {props.currData}
     </div>
     <div>
        {props.currData} ({props.higherOrLower})
      </div>
      {
           if (index === 0) return null;             
           const prevData = dataset[index - 1].value;
           const delta = (currData - prevData).toFixed(2);
           const deltaPercentage = ((currData / prevData - 1) * 100).toFixed(2);
           const currRevision = dataset[index].revision;
           const prevRevision = dataset[index - 1].revision;
           const hgURL = `https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=${prevRevision}&tochange=${currRevision}`;
           return <div> Δ {delta} ({deltaPercentage} %)</div>
    }
     <div>  <a href={hgURL}>{currRevision.slice(0, 12)}</a></div>

src/utils/chartJs/generateCustomTooltip.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import CustomTooltip from './CustomTooltip';
+
+const generateCustomTooltip = (canvas, tooltipModel, series) => {
```
klahnakoski: Use JSX to construct this `tooltip-container` div.  You should be able to set all the style properties to ensure it is properly positioned.


    <div id="tooltip-container' style={{top: positionY + tooltipModel.caretY - heightTooltip, ...}}>
        <CustomTooltip tooltipModel={tooltipModel} series={series} />
    </div>

the `style` property accepts a object; notice the double curly braces (one for code, the other for object creation)

then, you should be able to inline `CustomTooltip` to result in a single component

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,68 @@
...
+  },
+  tooltipKey: {
+    display: 'inline-block',
+    width: '10px',
+    height: '10px',
+    marginRight: '10px',
+  },
+};
+
+function Footer({ dataset, index, currData }) {
```
klahnakoski: In-lining is better:

```javascript
import React from 'react';

const styles = {
  tooltip: {
    background: 'rgba(0, 0, 0, .8)',
    color: 'white',
    borderRadius: '4px',
    pointerEvents: 'none',
  },
  tooltipKey: {
    display: 'inline-block',
    width: '10px',
    height: '10px',
    marginRight: '10px',
  },
};

function CustomTooltip({ classes, tooltipModel, series }) {
  if (tooltipModel.opacity === 0) return null;

  const firstPoint = tooltipModel.dataPoints[0];
  const labelColors = tooltipModel.labelColors[0];
  const { xLabel, index, datasetIndex } = firstPoint;
  const seriesLabel = series[datasetIndex].label;
  const higherIsBetter = !series[datasetIndex].meta.lower_is_better;
  const higherOrLower = higherIsBetter ? 'higher is better' : 'lower is better';
  const paddingStyle = {
    padding: `${tooltipModel.yPadding}px ${tooltipModel.xPadding}px`,
  };
  let footer = null;

  if (index > 0) {
    const [prev, curr] = series[datasetIndex].data.slice(index - 1);
    const delta = curr.value - prev.value;
    const deltaPercentage = (delta / prev.value) * 100;
    const hgURL = `https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=${
      prev.revision
    }&tochange=${curr.revision}`;

    footer = (
      <React.Fragment>
        <div>
          Δ {delta.toFixed(2)} ({deltaPercentage.toFixed(0)} %)
        </div>
        <div>
          <a href={hgURL}>{curr.revision.slice(0, 12)}</a>
        </div>
      </React.Fragment>
    );
  }

  return (
    <div className={classes.tooltip} style={paddingStyle}>
      <div>{xLabel}</div>
      <div>
        <span style={labelColors} className={classes.tooltipKey} />
        {seriesLabel}: {firstPoint.yLabel}
      </div>
      <div>
        {firstPoint.yLabel} ({higherOrLower})
      </div>
      {footer}
    </div>
  );
}

export default withStyles(styles)(CustomTooltip);

```

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,68 @@
...
+  const prevData = dataset[index - 1].value;
+  const prevRevision = dataset[index - 1].revision;
```
klahnakoski: instead of `prevData` and `prevRevision`

    const prev = dataset[index - 1]

and refer to `prev.value` and `prev.revision`

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,68 @@
...
+  const delta = (currData - prevData).toFixed(2);
```
klahnakoski: move the `toFixed(2)` to the JSX

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,68 @@
...
+    </React.Fragment>
+  );
+}
+
+function CustomTooltip({ classes, tooltipModel, series }) {
+  if (tooltipModel.opacity === 0) return null;
+
+  const { xLabel, index, datasetIndex } = tooltipModel.dataPoints[0];
+  const dataset = series[datasetIndex].data;
+  const seriesLabel = series[datasetIndex].label;
```
klahnakoski: use `const firstPoint = series[datasetIndex]` and use `firstPoint` where you can

src/utils/chartJs/CustomTooltip.jsx
klahnakoski: ...and use destructuring

    const [prev, curr] = series[datasetIndex].data.slice(index - 1);

klahnakoski:
Messing with the code more, I see many values getting pulled out of the tooltipModel. YOu can pull out most of the values in one line:

``javascript
  const {
    labelColors: [labelColors],
    dataPoints: [{ xLabel, yLabel, index, datasetIndex }],
  } = tooltipModel;
```

File: src/utils/chartJs/CustomTooltip.jsx [hunk 17]
```diff
@@ -0,0 +1,68 @@
...
+  const labelColors = tooltipModel.labelColors[0];
+  const currData = tooltipModel.dataPoints[0].yLabel;
+  const higherIsBetter = !series[datasetIndex].meta.lower_is_better;
+  const higherOrLower = higherIsBetter ? 'higher is better' : 'lower is better';
+  const paddingStyle = {
```
klahnakoski: Find out what this `paddingStyle` is and copy it into `styles` at the top of the file. Remove `paddingStyle`

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,68 @@
...
+  },
+};
+
+function Footer({ dataset, index, currData }) {
+  const prevData = dataset[index - 1].value;
+  const prevRevision = dataset[index - 1].revision;
+  const delta = (currData - prevData).toFixed(2);
+  const deltaPercentage = ((currData / prevData - 1) * 100).toFixed(2);
+  const currRevision = dataset[index].revision;
+  const hgURL = `https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=${prevRevision}&tochange=${currRevision}`;
```
klahnakoski: I solved my confusion from last night:  We do not need the previous revision to show `currRevision`. Here is an example URL:

    https://hg.mozilla.org/mozilla-central/pushloghtml?changeset=d326a9d5f77b

klahnakoski: ... so you can pull it out of the <React.Fragment>

klahnakoski:
Sorry, as this not-allowed-if-block-in-JSX continues to burn in my mind, I remember that any function can be turned into an expression:

    value = (() => { /*put any code here*/ return something; })();

We can use this pattern in JSX:

```javascript
  return (
    <div className={classes.tooltip} style={paddingStyle}>
      <div>{xLabel}</div>
      <div>
        <span style={labelColors} className={classes.tooltipKey} />
        {single.label}: {yLabel}
      </div>
      <div>
        {yLabel} ({higherOrLower})
      </div>
      {(() => {
        if (index > 0) {
          const [prev, curr] = single.data.slice(index - 1);
          const delta = curr.value - prev.value;
          const deltaPercentage = (delta / prev.value) * 100;
          const hgURL = `https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=${
            prev.revision
          }&tochange=${curr.revision}`;

          return (
            <React.Fragment>
              <div>
                Δ {delta.toFixed(2)} ({deltaPercentage.toFixed(0)} %)
              </div>
              <div>
                <a href={hgURL}>{curr.revision.slice(0, 12)}</a>
              </div>
            </React.Fragment>
          );
        }

        return null;
      })()}
    </div>
  );
```

src/utils/chartJs/CustomTooltip.jsx
2alin: Got it, just some things to point out:
- `tooltipModel.dataPoints[0]`doesn't refer to the first data point in the series, as you can check `tooltip`in [this gist](https://gist.github.com/2alin/2a1ef853d3dc9d251bb398b1f624c422).`tooltipModel.dataPoints` is an array with always one element that has the information of the current point, so I think it would be better to define it as `const currPoint = tooltipModels.dataPoints[0]`
- For `deltaPercentage`, I'm not sure that we should round it to an integer (`deltaPercentage.toFixed(0)`), there are many points in the chart (just checked it already) that gives a deltaPercentage of -0.38, 0.38, etc. I will round it to one decimal position, as it's set in the [treeherder graphs](https://treeherder.mozilla.org/perf.html#/graphs?series=mozilla-central,1775979,1,10&series=mozilla-central,1776166,1,10&series=mozilla-central,1776360,1,10&selected=mozilla-central,1775979,438078,733182131,10), but if you insist I'll set it as suggested then.
- 

About this:
> delay any formatting (eg .toFixed(2)) to the last possible line

I had it like that in a first version of another PR, but I was suggested to move it to how it was now. I suppose your reason is to keep the calculated value as precise as possible, and only trim it when it's displayed.

src/utils/chartJs/CustomTooltip.jsx
2alin: `paddingStyle`are options that are fetched from `tooltip.xPadding` and `tooltip.yPadding` chart options. By default it's value is 6px, if we set a fixed padding value to our custom tooltip (i.e., setting 6px in the `const styles = {...}`), then if in the future someone wants to set a padding inside `perfherderFormatter`, like this :
```jsx
// perherderFormatter.js
...
const generateInitialOptions = series => {
  ...
  return {
    ... 
    tooltips: {
      xPadding: '8px';
      yPadding: '4px';
      ...
      custom(tooltipModel) {...}
    }
  }
}
``` 
it **won't work** and will ignore that padding. Is there a way that we can set dynamically those values inside `const styles = {...} `? Because I can't see it. Or if you want to set fixed values as I mentioned before, we will lose a functionality in the chart options.

src/utils/chartJs/CustomTooltip.jsx
2alin: Yup, you are right, bravo!! :)
Should I use the short version of the hash(?)  (12 chars) to generate the link or the long one?

2alin:
> Essentially, we define a function, and run it right away:
> 
> ```js
>     (() => {/*code*/})()
> ```

Awesome!! and pretty handy. I think I saw it once somewhere, at least the syntax looks familiar to me, but I didn't have the opportunity to use it. 

Thank you for your review @klahnakoski, I have commited almost all your suggestions in my local repo, and I'll wait for you to answer some concerns I wrote up there before pushing it to remote.

src/utils/chartJs/CustomTooltip.jsx
klahnakoski: Yes, I am aware these types of styling "features" will be lost; that is the side effect of having a CustomTooltip.  

There should be one place that has all the styling information, and it should be the `const styles` at the top of the file.  Setting stylings in another place, like the `tooltips` substructure, runs the risk of being confusing.  Suppose someone does set `xPadding: '4px'`, and I try to set the `padding` in the `const styles`: I am in for some pain as I try to figure out why my styling is not working.

src/utils/chartJs/CustomTooltip.jsx
klahnakoski: Longer revision is better.

src/utils/chartJs/CustomTooltip.jsx
klahnakoski: * On the subject of `const currPoint = tooltipModels.dataPoints[0]`: Feel free to pick better names.
* For `deltaPercentage`, there are already 2 significant digits to the left of the decimal point.  Parts of a percent is unreasonably precise.  But matching Treeherder is fine too.
* On ` .toFixed(2)`, it is translating to a humane, but inaccurate, format. You will notice I can use the `delta` to calculate the percentage because I did not loose accuracy. We also keep the translation code together; in this case all in the JSX.

2alin:
Now, I was going to implement the suggestions you gave me days ago about `generateCustomTooltip` (i.e. the custom tooltip container). And I got an idea that I think you will freaking love it, and it changes completely how we are injecting the tooltip container in the DOM. I already made some tests and it's doable, but there are important changes to be applied in other files, so I'll implement this idea in another branch, show it to you, and if you like it, then I'll merge it into this one. Just gave me some days, because I'm a little busy as I explained before.

src/utils/chartJs/generateCustomTooltip.jsx
2alin: For this suggestion, please check comment at the bottom of the main thread.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,74 @@
...
+  const currPoint = tooltipModel.dataPoints[0];
+  const labelColors = tooltipModel.labelColors[0];
+  const { index } = currPoint;
+  const currSeries = series[currPoint.datasetIndex];
+  const higherIsBetter = !currSeries.meta.lower_is_better;
+  const higherOrLower = higherIsBetter ? 'higher is better' : 'lower is better';
+  const curr = currSeries.data[index];
+  const hgURL = `https://hg.mozilla.org/mozilla-central/pushloghtml?changeset=${
+    curr.revision
+  }`;
```
klahnakoski: use this

```javascript
  const hgURL = `https://hg.mozilla.org/mozilla-central/pushloghtml?${Object2URL({changeset: curr.revision})}`;
```

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,74 @@
...
+  const jobURL = `https://treeherder.mozilla.org/#/jobs?repo=mozilla-central&revision=${
+    curr.revision
+  }&selectedJob=${curr.job_id}&group_state=expanded`;
```
klahnakoski: use this

```javascript
const jobURL = `https://treeherder.mozilla.org/#/jobs?${
  Object2URL({
    repo: "mozilla-central",
    revision: curr.revision,
    selectedJob: curr.job_id,
    group_state: "expanded"
})}`
```

generally, we would like to avoid string concatenation as much as possible

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,74 @@
...
+          <div>
+            Δ {delta.toFixed(2)} ({deltaPercentage.toFixed(1)} %)
+          </div>
+        );
+      })()}
+      <div>
+        <a href={hgURL} target="_blank" rel="noopener noreferrer">
+          {curr.revision.slice(0, 12)}
+        </a>
+        {` `}(
```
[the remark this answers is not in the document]
klahnakoski: I am not sure either.  looks fine

klahnakoski:
@2alin So, what's the status of this?  If you are not able to continue work, then I will take over and bring to the finish line.

2alin:
@klahnakoski I'll have something for you to check out tonight.

2alin:
@klahnakoski 
And here it is: #318
Please check it. Your last suggestions in this branch are still on hold. I'll implement them once we decide which branch to follow.

src/utils/chartJs/CustomTooltip.jsx
2alin: I'm getting `'Object2URL' is not defined`, should I define that function with the purpose that you are trying to get?

PR #319, opened 2019-02-24
fix handle click over legend to hide/show series

### LINKED ISSUE #229: Add ability to hide/show series on PerfherderGraphContainer
This is a screenshot of what you can currently do in Perfherder:
image

The checkbox allows removing data series as well as showing/hiding them.

2alin:
Fix #229 
@klahnakoski @armenzg 
I noticed that by default ChartJS shipped their charts with the ability to hide/show series, but somehow this was being blocked/changed in our project. After tracking down the issue, it was `registerTooltip.jsx`the culprit. I only added the case when the click event happened outside of a data point.

As you can see in [1], it looks nice. 

Now, I think we should change the minimum axes range, because in the Perfherder graphs, the data points are far from 0, and even after hiding some series, the plot doesn't look as good as it could be, but that's a different issue to solve :)

[1]
ezgif-4-f270f71a420e


2alin:
Thank you for merging it :)

PR #329, opened 2019-03-04
add to common libs

klahnakoski:


klahnakoski:
@helfi92 

This PR has no effect, I am adding library functions from my old javascript dashboards, which I will be using later.  In the meantime, I added tests, and the coverage should indicate the code is reasonably bug-free.

klahnakoski:
@armenzg  I do not know how busy you are catching up:  These files do almost nothing, but are required for the next set of PRs.  Notice the coverage.

klahnakoski:
@armenzg  it is squashed, but the merge history is all lost.

PR #338, opened 2019-03-08
More mobile

klahnakoski:
based on the `bench` branch


src/quantum/config.js
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable max-len */
+/* eslint-disable */
```
armenzg: Could you please choose specific rules to disable rather than all rules?

src/quantum/config.js
```diff
@@ -161,10 +161,31 @@ const SUITES = {
     ['Chromium', 'Tp6: Microsoft', 'raptor-tp6-microsoft-chrome'],
     ['Chromium', 'Tp6: Reddit', 'raptor-tp6-reddit-chrome'],
 
-    ['geckoview', 'Tp6 mobile: Facebook', 'raptor-tp6m-facebook-geckoview'],
+
     ['geckoview', 'Tp6 mobile: Amazon', 'raptor-tp6m-amazon-geckoview'],
-    ['geckoview', 'Tp6 mobile: YouTube', 'raptor-tp6m-youtube-geckoview'],
+    ['geckoview', 'Tp6 mobile: Amazon Search', "raptor-tp6m-amazon-search-geckoview"],
```
armenzg: I can speak of the correctness of this block.
It seems reasonable from a quick glance.

src/quantum/config.js
klahnakoski: I was unable to determine what rules continue to rewrite the lines of this file.

PR #340, opened 2019-03-08
Rename to Vector

klahnakoski:
rename `frum` to `vector`

### DIFF
--- a/src/components/SummaryTable/index.jsx
+++ b/src/components/SummaryTable/index.jsx
@@ -3,11 +3,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import StatusWidget from '../StatusWidget';
 import CONFIG from '../../utils/nimbledroid/config';
-import { frum } from '../../vendor/queryOps';
+import { vector } from '../../vendor/vectors';
 
 const SummaryTable = ({ content = [], header }) => {
   const compareName = CONFIG.packageIdLabels[CONFIG.baseProduct];
-  const compareColumn = frum(header).findIndex(name => name === compareName);
+  const compareColumn = vector(header).findIndex(name => name === compareName);
 
   return (
     <table className="summary-table">
--- a/src/quantum/config.js
+++ b/src/quantum/config.js
@@ -1,6 +1,6 @@
 /* eslint-disable */
 
-import { frum } from '../vendor/queryOps';
+import { vector } from '../vendor/vectors';
 import Data from '../vendor/Data';
 
 const CONFIG = {
@@ -189,7 +189,7 @@ const SUITES = {
   ],
 };
 // ALL PAGE COMBINATIONS
-const TP6_PAGES = frum(SUITES.data)
+const TP6_PAGES = vector(SUITES.data)
   .map(row => Data.zip(SUITES.header, row))
   .join('browser', PLATFORMS, 'browser');
 const TP6M_PAGES = TP6_PAGES.where({ browser: 'geckoview' });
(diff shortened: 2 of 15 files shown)


klahnakoski:
@sarah-clements  how about `vector`?  is that a nicer name?

sarah-clements:
> @sarah-clements how about `vector`? is that a nicer name?

I haven't heard of vectors being used in Javascript to describe dynamic arrays (if that's why you're using it). I still think naming the method by describing what it does will make it clearer (and somewhat self-documenting), even if it ends up a little verbose. Maybe @armenzg has an opinion?

armenzg:
I 100% agree with @sarah-clements 
Are we talking about an array of arrays? Would `arrayOfArrays` help?

klahnakoski:
@sarah-clements @armenzg 

The word vector comes from [linear algebra](https://en.wikipedia.org/wiki/Linear_algebra). It is commonly used as synonym for Array, Java [has Vectors](https://docs.oracle.com/javase/7/docs/api/java/util/Vector.html) which are like Javascript Arrays.  This library provides a number of vector operations that can be found in [array programming languages](https://en.wikipedia.org/wiki/Array_programming).

klahnakoski:
Since @armenzg was away, here is my main justification: https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/291#issuecomment-462057078

sarah-clements:
> @sarah-clements @armenzg
> 
> The word vector comes from [linear algebra](https://en.wikipedia.org/wiki/Linear_algebra). It is commonly used as synonym for Array, Java [has Vectors](https://docs.oracle.com/javase/7/docs/api/java/util/Vector.html) which are like Javascript Arrays. This library provides a number of vector operations that can be found in [array programming languages](https://en.wikipedia.org/wiki/Array_programming).

Since you're asking for an opinion, I'll reiterate that calling it `vectors` (translating to `arrays`) doesn't actually describe to someone looking at that code for the first time what the method is doing. If I recall, it's arg is an array and it's creating an instance from that array from which to chain. I think we can put together a useful, readable name from that description (if correct), that others who didn't write that library will be able to understand. :)

klahnakoski:
Well, `frum` does nothing useful on its own, it is an [adapter](https://en.wikipedia.org/wiki/Adapter_pattern) or [facade](https://en.wikipedia.org/wiki/Facade_pattern).

> doesn't actually describe to someone looking at that code for the first time what the method is doing.

Does it help that `frum` will do nothing unless it is chained to another method?  We will never see `frum` on its own, which is why I never gave much priority to being descriptive; the subsequent functions are descriptive.

> If I recall, it's arg is an array and it's creating an instance from that array from which to chain. I think we can put together a useful,readable name from that description. 

Lodash uses `chain`.

And, thank you for taking the time to expand on where I am unclear.

armenzg:
> @sarah-clements @armenzg
> 
> The word vector comes from [linear algebra](https://en.wikipedia.org/wiki/Linear_algebra). It is commonly used as synonym for Array, Java [has Vectors](https://docs.oracle.com/javase/7/docs/api/java/util/Vector.html) which are like Javascript Arrays. This library provides a number of vector operations that can be found in [array programming languages](https://en.wikipedia.org/wiki/Array_programming).

Thanks for the reminder but a lot of our community does not come with Math background. Naming easier to digest is always benefitial.

This is one of the concerns I shared back in December when adding this approach to manipulating data from config files. It only makes things harder to understand and even harder barrier to access from our community.

PR #341, opened 2019-03-08
HOC for Errors (version 3)

klahnakoski:
This is based on top of the `more-mobile` branch so only the net difference is seen


src/containers/BugzillaGraph/index.jsx
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import withErrorBoundary from '../../hocs/withErrorBoundary';
+import { withErrorBoundary } from '../../vendor/errors';
```
armenzg: I must be too out of the groove but I can't figure out why this change would be needed.
I would have assumed that using `import foo from '/path/to/foo'` would work.

src/vendor/errors.jsx
```diff
@@ -1,154 +1,255 @@
...
+if (isNode) {
+  //   Error
+  //       at Function.Object.<anonymous>.Log.error (C:\Users\kyle\code\firefox-health-dashboard\src\vendor\errors.jsx:174:9)
+  //       at Object.error (C:\Users\kyle\code\firefox-health-dashboard\test\vendor\errors.test.js:9:16)
+  //       at Object.asyncJestTest (C:\Users\kyle\code\firefox-health-dashboard\node_modules\jest-jasmine2\build\jasmine_async.js:108:37)
+  //       at resolve (C:\Users\kyle\code\firefox-health-dashboard\node_modules\jest-jasmine2\build\queue_runner.js:56:12)
+  //       at new Promise (<anonymous>)
+  //       at mapper (C:\Users\kyle\code\firefox-health-dashboard\node_modules\jest-jasmine2\build\queue_runner.js:43:19)
+  //       at promise.then (C:\Users\kyle\code\firefox-health-dashboard\node_modules\jest-jasmine2\build\queue_runner.js:87:41)
+  //       at process.internalTickCallback (internal/process/next_tick.js:77:7)
```
armenzg: Do you need this commented block?

src/vendor/errors.jsx
```diff
@@ -1,154 +1,255 @@
...
-    const error = Exception.wrap(err, info);
-
-    this.setState({ error });
-
-    Log.warning(err);
-  }
-
-  render() {
-    const { error } = this.state;
+const withErrorBoundary = WrappedComponent => {
```
armenzg: Why is `withErrorBoundary` in this module instead of its own?

src/vendor/errors.jsx
```diff
@@ -1,154 +1,255 @@
 /* eslint-disable react/no-multi-comp */
+/* eslint-disable max-len */
 import React from 'react';
-import PropTypes from 'prop-types';
-import { missing } from './utils';
+import ErrorPanel from '@mozilla-frontend-infra/components/ErrorPanel';
+import { coalesce, exists, isChrome, isNode, isString, missing } from './utils';
```
armenzg: How are all of these changes related to wrapping components with an HOC? (up to line 200)

src/vendor/errors.jsx
armenzg: I'm also noticing a lot of new classes in here that I don't know what value they're adding (e.g. Exception.wrap)and we're lossing some important feature like reporting to Sentry [1] or reporting depending if we're on production vs development [2]

I would need a strong case to take changes that would make us loose this.

[1] https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/13c943fb3c788647eaefe138e3f71368669cc930/src/components/genericErrorBoundary.jsx#L13
[2] https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/13c943fb3c788647eaefe138e3f71368669cc930/src/hocs/withErrorBoundary.jsx#L6

armenzg: Also this wrapping things are causing them to change the normal behaviour of the console logging.
For instance, this error should not be an info message (it should show red):
image

Can you please move all of this logic affecting console logging to be moved into another PR so we can just focus on error boundaries? Or better than that, file an issue so we can understand why these changes would be needed. I would not want you to spend time creating another PR if there's no clear value to be gained.

src/vendor/utils.js
```diff
@@ -119,7 +120,20 @@ function concatField(...many) {
...
+const isBrowser = !isNode;
+const isChrome =
+  isBrowser && navigator.userAgent.toLowerCase().indexOf('chrome') > -1;
+const isFirefox = isBrowser && !isChrome;
+
 export {
+  isNode,
+  isBrowser,
+  isChrome,
+  isFirefox,
```
armenzg: How are any of these related to error boundaries?

.gitignore
```diff
@@ -8,3 +8,7 @@ dist
 build
 // Coverage (nyc)
 coverage
+/src/vendor/components
+/src/vendor/theme.js
+/src/vendor/styleguide
+/src/vendor/utils/palette.js
```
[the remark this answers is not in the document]
klahnakoski: I wanted to discuss adding them with you first, and it is not part of this PR

src/components/ChartJsWrapper/index.jsx
```diff
@@ -33,31 +29,29 @@ const ChartJsWrapper = ({
...
-  if (data && missingDataError) {
-    showError = data.datasets.every(dataset => {
-      const latestDataDate = new Date(dataset.data[dataset.data.length - 1].x);
-      const currentDate = new Date(); // get current date
-      const timeDifference = Math.abs(
-        currentDate.getTime() - latestDataDate.getTime()
-      );
-      const daysDifference = Math.ceil(timeDifference / (1000 * 3600 * 24));
+  if (data) {
+    if (
```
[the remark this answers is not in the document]
klahnakoski: Yes, this was made complicated so it would snot complain if at least one series has recent data.  Your suggestion of showing the chart, with the error can simplify this logic.

src/containers/BugzillaGraph/index.jsx
klahnakoski: When I moved `../hocs/withErrorBoundary` to `../vendor/errors` I added more exportable items.

src/vendor/errors.jsx
klahnakoski: I would like to keep this:  Every regex should have examples of what it matches.

package.json
```diff
@@ -8,7 +8,7 @@
     "start:local": "BACKEND=http://localhost:3000 webpack-dev-server --mode development",
     "test": "jest",
     "lint": "eslint --cache --ext mjs,jsx,js --format codeframe \".*.js\" src test",
-    "precommit": "yarn lint --fix && lint-staged",
+    "precommit": "lint-staged",
```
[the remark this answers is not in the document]
klahnakoski: I added it earlier, but it does not work:  The pre-commit hook only runs after the files are staged, so the `lint --fix` never get committed:  After commit, we must still commit the fixes.

src/components/ChartJsWrapper/index.jsx
```diff
@@ -20,10 +20,6 @@ const styles = {
     backgroundColor: 'black',
     padding: '.3rem .3rem .3rem .3rem',
   },
-  errorPanel: {
-    margin: '26px auto 40px',
-    width: '70%',
-  },
 };
 const ChartJsWrapper = ({
```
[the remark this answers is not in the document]
klahnakoski: OK, but that blows my mind, and I am going to forget that.

klahnakoski:
@armenzg  Before more review, please help with this css problem:  The big red box is deliberate while I try to figure out how to show a border, any border, using CSS.  Here is [the style object](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/341/files#diff-d980d253af9020948727c19035dd089bR24) here is [the preview](https://deploy-preview-341--firefox-health-dashboard.netlify.com/android). Open debugger, click on the nimbledroid graph, mess with the CSS properties. What special combination gets a border to show. 

image

I have tried borders in pixels, and in rem. I have tried changing box-sizing, margins and padding. Nothing works. I have this type of problem often with CSS: Assigning a style to an object simply does not work.  What am I missing?

klahnakoski:
NM, I knew that once I sent you a message, it would reveal itself: Of course, you must set the border-style to "solid", because "none" is the default.

klahnakoski:
@there are a few other tiny changes, but to minimize pushes, I will leave them to after your review.

src/index.jsx
```diff
@@ -3,28 +3,66 @@ import React from 'react';
...
+
+    return this.props.children;
+  }
+}
 const root = document.getElementById('root');
 const load = () =>
   render(
     <AppContainer>
-      <GenericErrorBoundary critical>
+      <GlobalErrorBoundary>
```
armenzg: I think the only difference with `withErrorBoundary` is that it does not show the error message (e.g. `<ErrorMessage error={error}>{super.render()}</ErrorMessage>`) but a customized UI.

I wonder if we can do something like this in routes.jsx:
```
const ComponentWithMessage = (
  <div>File a precious issue!</div>
);

export default withErrorBoundary(Routes, ComponentWithMessage);
```

And then make this change to the signature:
```
const withErrorBoundary = (WrappedComponent, ShowOnError = null) => {
```

This is not that important but I thought we might be able having two different classes to handle error boundaries.

src/quantum/index.jsx
```diff
@@ -30,20 +29,20 @@ export default class QuantumIndex extends React.Component {
       match: { params },
     } = this.props;
     const urlParams = URL2Object(location.search);
-    const bits = urlParams.bits || params.bits;
+    const bits = urlParams.bits || Number.parseInt(params.bits, 10);
```
armenzg: I know it's very tempting but in the future try not to have drive-by fixes to make the PRs extra clear.
If it helps `git add --patch` has done wonders to me to separate changes from the same file.

src/quantum/index.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { parse } from 'query-string';
 import Grid from '@material-ui/core/Grid/Grid';
 import DashboardPage from '../components/DashboardPage';
-import Perfherder from './perfherder';
+import PerfherderWidget from './perfherder';
```
armenzg: In the future, don't spend your time investing on this component.
We need to switch these two graphs to using `PerfherderGraphContainer` and we can just kill the old code:
image

src/components/ChartJsWrapper/index.jsx
```diff
@@ -33,50 +33,66 @@ const ChartJsWrapper = ({
...
-  return data ? (
-    <div className={classes.chartContainer}>
-      {showError && (
-        <CriticalErrorMessage
-          className={classes.errorPanel}
-          error="This item has been missing data for at least 3 days."
+    if (error) {
+      return (
+        <div className={classes.chartContainer}>
+          <ErrorMessage error={error}>
```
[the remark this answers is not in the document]
klahnakoski: I added `pointer-events: none` so the mouse events get through. Covering the axis was deliberate; I did not want to make the element bigger and possibly mess with the layout.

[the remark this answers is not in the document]
klahnakoski: I messed with opacity.

src/vendor/errors.jsx
```diff
@@ -1,154 +1,115 @@
...
+    const {
+      error,
+      classes: { container, frame, message },
+      children,
+    } = this.props;
+
+    return (
+      <div className={container}>
+        {children}
+        <div className={frame} />
```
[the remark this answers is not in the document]
klahnakoski: the `frame` is the red outline around the component with the error

src/vendor/errors.jsx
```diff
@@ -1,154 +1,115 @@
...
-class ErrorMessage extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {};
+const ErrorMessage = withStyles(styles)(RawErrorMessage);
+const withErrorBoundary = WrappedComponent => {
+  if (
+    WrappedComponent.displayName &&
+    WrappedComponent.displayName.startsWith('WithStyles')
+  ) {
```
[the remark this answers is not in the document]
klahnakoski: `componentDidMount()` is a function that returns `undefined`.  `withStyles` returns a component that does just that. 

This is problem because we use `componentDidMount()` to launch any number of promise chains that may fail sometime.  Our components use `async componentDidMount()` which is just a function the returns a promise; React ignores the returned promise, but that does not stop the promise chain from running.

The `withErrorBoundary` is creating a class that uses the promise we are returning from `async componentDidMount()` and ensuring raised errors are properly caught and recorded.

Therefore, 

    export default withErrorBoundary(withStyles(styles)(PerfherderGraphContainer));

is not allowed

src/vendor/errors.jsx
```diff
@@ -1,154 +1,115 @@
...
+        return <ErrorMessage error={error} />;
+      }
     }
   }
 
   return ErrorBoundary;
 };
 
-export { Exception, withErrorBoundary, ErrorMessage, Log };
+export { withErrorBoundary, ErrorMessage };
```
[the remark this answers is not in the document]
klahnakoski: ErrorMessage should be private, but I used it for the `days>3` check. I am not sure that check should be in the render method, but that is problem for later.  Until then we have some leakage.

src/quantum/index.jsx
klahnakoski: I noticed this problem, It thought it was just something that had to be fixed to get the pages working.  Turns out that this is an existing problem: https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/348

src/vendor/errors.jsx
```diff
@@ -1,154 +1,115 @@
...
+  ) {
+    throw new Error(
+      'Can not wrap WithStyles because componentDidMount() returns undefined'
+    );
   }
 
-  componentDidCatch(err, info) {
-    const error = Exception.wrap(err, info);
+  class ErrorBoundary extends WrappedComponent {
+    constructor(...props) {
```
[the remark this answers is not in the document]
klahnakoski: Yes, some components take multiple properties.

src/vendor/errors.jsx
```diff
@@ -1,154 +1,115 @@
...
+      super(...props);
 
-    this.setState({ error });
-
-    Log.warning(err);
-  }
-
-  render() {
-    const { error } = this.state;
+      if (missing(this.state)) {
```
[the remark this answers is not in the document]
klahnakoski: Some components, one defined using just functions, do not set a state object.  Here we make a stateobject, if one is not already made.

klahnakoski:
@armenzg   Another round of review please!

klahnakoski:
@armenzg Another round of review please!

.gitignore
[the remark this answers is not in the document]
armenzg: Please remove these as they're not by-products of the tools we use (e.g. a build/ directory).

src/components/ChartJsWrapper/index.jsx
```diff
@@ -33,50 +34,70 @@ const ChartJsWrapper = ({
...
+    }
+
+    let error = null;
 
-  if (data && missingDataError) {
-    showError = data.datasets.every(dataset => {
-      const latestDataDate = new Date(dataset.data[dataset.data.length - 1].x);
+    data.datasets.forEach(dataset => {
+      const latestDataDate = new Date(
+        frum(dataset.data)
```
armenzg: I know you love your chaining library but it is not adding a lot more value and it forces new users (and old users like me) to have to learn a new way of doing things rather than using standard data manipulations practices.

I believe there's more value to not switch to it for the sake of contributions.

src/components/ChartJsWrapper/index.jsx
```diff
@@ -33,50 +34,70 @@ const ChartJsWrapper = ({
...
-        textAlign: 'center',
-        width: spinnerSize,
-      }}>
-      <CircularProgress />
-    </div>
-  );
-};
+      </div>
+    );
+  })();
```
armenzg: Why did you turn it into a function call?

src/components/ChartJsWrapper/index.jsx
```diff
@@ -33,50 +34,70 @@ const ChartJsWrapper = ({
...
   chartHeight,
   spinnerSize,
-  missingDataError = false,
-}) => {
-  let showError;
+}) =>
+  (() => {
+    if (!data) {
+      return (
+        <div
```
armenzg: The problem of moving the rendering of the NO DATA case scenario up here is that there are now two different locations to determine what does this method render.
I would prefer that the two were together. Your call.

src/components/ChartJsWrapper/index.jsx
```diff
@@ -33,50 +34,70 @@ const ChartJsWrapper = ({
...
-  return data ? (
-    <div className={classes.chartContainer}>
-      {showError && (
-        <CriticalErrorMessage
-          className={classes.errorPanel}
-          error="This item has been missing data for at least 3 days."
+    if (error) {
+      return (
+        <div className={classes.chartContainer}>
+          <ErrorMessage error={error}>
```
armenzg: Lines 77 to 83 are duplicated of 91 to 97.

You could merge the two by doing this (notice the line for errorMessage):
```javascript
    return (
      <div className={classes.chartContainer}>
        {title && <h2>{title}</h2>}
        <Chart
          type={type}
          data={data}
          height={chartHeight}
          options={generateOptions(options)}
        />
        {error && <ErrorMessage error={error}>}
      </div>
    );
```

You don't need to wrap the whole graph to make an error message obvious (see comparison):
image

src/vendor/errors.jsx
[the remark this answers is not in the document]
armenzg: I mentioned it on another conversation above.
The code and UI will be easier if you don't try to wrap things.
I will leave it up to you to make the call given the input.

src/components/ChartJsWrapper/index.jsx
klahnakoski: This is a legit concern, but what's happening here is a side effect of showing errors in general: **Now we get to see all the errors!**  There are two more problems, which we can solve in another set of PRs:

1.  This particular component is doing too much: this code redundancy will disappear once it is fixed: https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/351
2. The type of error may dictate how best to reveal it to the user: this is partly solved with your earlier suggestion of `withErrorBoundary(specificErrorCompoenent)(wrappedCompoenent)`, which we can also do later.

src/components/ChartJsWrapper/index.jsx
klahnakoski: The `(() => {...})()` pattern turns a block of code (`...`) into an expression.  This allows us to make the code more readable:

Instead of trinary expressions and Boolean `&&` tricks:

```javascript
data ? ( showError && <error> ) <data> : <spinner>
```

we can say 

```javascript
(()=>{
    if (showError) return <error>;
    if (!data) return <spinner>;
    return <data>;
})();
```

which is much nicer when the `<components>` are complicated

src/components/ChartJsWrapper/index.jsx
klahnakoski: I find this clearer.  When `!data` we show `<CircularProgress>`

The code changed to expose the fact it is actually has three distinct states, and shows three distinct compoenents:  https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/351

src/components/ChartJsWrapper/index.jsx
klahnakoski: For the future, when you suggest "standard data manipulations practices", please include the code you would like to see.  I do not know what you mean by "standard".

Instead of this

      const latestDataDate = new Date(
        frum(dataset.data)
          .sort('x')
          .last().x
      );

use this

      const sortedTimestamps = dataset.data.map(({x}) => x).sort();
      const latestDataDate = new Date(sortedTimestamps[sortedTimestamps.length-1]);

or this

      const latestDataDate = new Date(
        dataset.data
          .map(({x}) => x)
          .sort()
          .reverse()[0]
      );

or this

      const latestDataDate = new Date(Math.max(dataset.data.map(({ x }) => x)));

which is the nicest.

PR #342, opened 2019-03-08
Disable browser lint warnings

klahnakoski:


klahnakoski:
As per the same documentation: 

> To specify environments in a configuration file, use the env key and specify which environments you want to enable by setting each to true. 

To me that reads as "enable, or disable, eslint for particular environments".

With respect to 

> Each environment brings with it a certain set of predefined global variables

I am guessing that someone that knows too much and is exposing the underlying implementation.

klahnakoski:
@armenzg well, running this again, it does not prevent lint problems showing the browser.

PR #348, opened 2019-03-11
Handle bits url parameter better

klahnakoski:


klahnakoski:
@armenzg This PR is ready for review, it fixes a bug in the handling of the "bits" URL parameter.  The important part are the tests:  I want to test "https://health.graphics/quantum/tp6?test=loadtime&bits=64" shows the correct set of tests.  The default (no parameters), and injecting parsed parameters, are also tests, which always worked.

test/views/QuantumTP61.test.jsx
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable linebreak-style */
+import renderer from 'react-test-renderer';
+import { MemoryRouter } from 'react-router-dom';
```
armenzg: Fantastic. Thank you for finding this.

PR #355, opened 2019-03-14
WIP: Add structured logging, and chained exceptions

klahnakoski:


armenzg:
What are you trying to solve?
This change will cause more code strongly tied to such module.

There are some code conflicts to be solved.

armenzg:
If there's a clear value, I would prefer we use a 3rd party library rather than have to burden ourselves with maintaining it.

klahnakoski:
@armenzg 

Here are the features I am looking for:

* structured logging - stop using template literals. so logs are machine readable 
* chain exceptions - so high level reasons can be assigned to low level causes
* send detailed errors to console - so the UI can show something simple, while the console can be used to diagnose (or anything downstream)
* switch error mitigation based on reason for error - switch based on `reason`

The API I have been using has three functions:

* `Log.note(reason, data)` - simple log message
* `Log.warning(reason, data, cause)` - Send error to console, no longer an error
* `Log.error(reason, data, cause)` - throw an error

I am agnostic about the names, or the actual library that implements this functionality.

armenzg:
Hi @klahnakoski 
I've looked at the [log module](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/f3b9c013243bf2bde8f02b12cba00ccb0808162f/src/vendor/logs.js) and as a team, I would vote against taking this change. It's prone to breakage due to regexes (Chrome or Firefox formatting changes), it adds one more module to have to maintain, it is a low-level module and it adds another different way for contributors to learn.

For changes like this, I would also encourage using the mozilla.tools newsgroup and decide as a community.

PR #357, opened 2019-03-19
Use query-string to convert objects to url paramters, and back

klahnakoski:
Add more tests, get tests to pass

### DIFF
--- a/src/quantum/index.jsx
+++ b/src/quantum/index.jsx
@@ -5,7 +5,7 @@ import Grid from '@material-ui/core/Grid/Grid';
 import DashboardPage from '../components/DashboardPage';
 import Perfherder from './perfherder';
 import { frum, toPairs } from '../vendor/queryOps';
-import { URL2Object } from '../vendor/convert';
+import { FromQueryString } from '../vendor/convert';
 import TelemetryContainer from '../telemetry/graph';
 import {
   quantum32QueryParams,
@@ -29,7 +29,7 @@ export default class QuantumIndex extends React.Component {
       location,
       match: { params },
     } = this.props;
-    const urlParams = URL2Object(location.search);
+    const urlParams = FromQueryString(location.search);
     const bits = urlParams.bits || Number.parseInt(params.bits, 10);
     const quantumQueryParams =
       bits === 32 ? quantum32QueryParams : quantum64QueryParams;
--- a/src/utils/perfherder/subbenchmarks.jsx
+++ b/src/utils/perfherder/subbenchmarks.jsx
@@ -1,7 +1,7 @@
 /* global fetch */
 import percentile from 'aggregatejs/percentile';
 import { frum, toPairs } from '../../vendor/queryOps';
-import { Object2URL } from '../../vendor/convert';
+import { ToQueryString } from '../../vendor/convert';
 
 const TREEHERDER = 'https://treeherder.mozilla.org';
 const PROJECT = 'mozilla-central';
@@ -57,7 +57,7 @@ const dataUrl = ({
       .toArray(),
   };
 
-  return `${TREEHERDER}/api/project/${project}/performance/data/?${Object2URL(
+  return `${TREEHERDER}/api/project/${project}/performance/data/?${ToQueryString(
     param
   )}`;
 };
--- a/src/vendor/convert.js
+++ b/src/vendor/convert.js
@@ -1,29 +1,20 @@
 import { parse } from 'query-string';
 import { frum, leaves, length, toPairs } from './queryOps';
 import { Log } from './errors';
-import {
-  isArray,
-  isFunction,
-  isNumeric,
-  isObject,
-  toArray,
-  exists,
-} from './utils';
+import { exists, isArray, isFunction, isObject, isString, toArray, } from './utils';
 import strings from './strings';
 
 function FromQueryString(query) {
   const decode = v => {
     if (isArray(v)) return v.map(decode);
 
-    if (v === null || v === 'true' || v === '') return true;
+    if (v === null || v === '') return true;
 
-    if (v === 'false') return false;
-
-    if (v === 'null') return null;
-
-    if (isNumeric(v)) return Number.parseFloat(v);
-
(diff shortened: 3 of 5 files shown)


klahnakoski:
The important part of this PR is the test suite: https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/357/files#diff-4494bd5de256ef856595f7ba56b0aae6  

It is testing some "reversible", and none-reversible query parameter translations.

klahnakoski:
Yes, I `Object2URL` and `URL2Object` are already in use, and the code that uses it makes some assumptions about encoding/decoding.  So, it was important to add test cases.  Which then turned into more tests for corner cases.

I did look into `query-string` but it does not map query parameters to objects as I would like, here are some ways it fails:

```
    { a: false } ⇔ 'a=false'
    { a: true } ⇔ 'a'
    { a: ' ' } ⇔ 'a=+'
    { a: '  ' } ⇔ 'a=++'
    { a: { test: 42 } } ⇔ 'a.test=42'
    { a: [1, 2, 3] } ⇔ 'a=1&a=2&a=3
```
also `http://example.com?a` [does not seem to parse correctly](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/357/files#diff-4494bd5de256ef856595f7ba56b0aae6R127):

```javascript
const {a}=parse("a");
if (a) {
    console.log("good")
} else {
    console.log("bad")
}
```

This is why I mentioned looking at the test cases. Are these test cases "wrong"?  If I want to convert `?a=42` into `{a: 42}`  where would that logic belong?  Should I make some utility methods to do that for me?

armenzg:
Just to wrap up, please try to see if `qs` would help and try to go for the regular test cases rather than everything that all the possiblities that we might not need to ever support.

Thanks for taking the time to discuss it with me and your patience on giving me clear examples.

klahnakoski:
@helfi92 and @djmitche 

I am looking to establish a "standard" library for frontend-infra.  I can not find one that meets my needs, (which I expressed in test cases) so I wrote one.  Maybe either of you know a library that does this, or know how to think of this problem in a different way.

I am not suborn on all the corner cases, but it would be nice to keep the query<->object mapping bijective as much as possible; having the numbers parsed is step in that direction.

Query strings are for servers, and this react app is acting as a server in it's own way; which is why we need to define query string parsing behaviour.  In general, talking to other servers, I would imagine each has their own idiosyncrasies that require their own query encoding, and belongs in whatever code is responsible for making those requests.

This is not just for query strings, but also fragment strings (in the case of static-served apps).

djmitche:
I may be out of my depth here.  I have a few quesetions, though:
 * What are all these functions in queryOps?  What does "frum" mean?  "leaves"?  If that library is a part of the frontend infra catalog, then I can't find it, nor the documentation for it.  And not understanding that makes it difficult for me to understand the code.
 * Why are patches occurring in a "vendor" directory?  Is that with the intent to upstream these patches to their source?
 * Why are the query strings for this particular application unique and requiring handling of type conversion?  The `qs` library parses the "80%" of queries, and using that has the advantage of forcing the app to use query strings in a way that is not surprising to others.  Is there a possibility to simplify the formatting of the query strings wherever they are generated, so that a thin wrapper over `qs` could parse them?
 * Speaking of surprise, type inference on untrusted data could lead to unexpected behavior, which could lead to security issues.  Obviously not in a frontend-only app, but if this library were used on the backend, then a parameter where a user chooses to write "0" or "null" instead of the expected string might result in a falsy value where one is not expected.
 * How would a user pass the string "null" using this library?
 * Why not use an existing type-preserving and well-tested encoding like JSON?

klahnakoski:
> What are all these functions in queryOps?  What does "frum" mean?  "leaves"?  If that library is a part of the frontend infra catalog, then I can't find it, nor the documentation for it.  And not understanding that makes it difficult for me to understand the code.

That is part of another conversation; finding a name for wrapping an array for chained vector operations, just like Lodash, but Lodash is too big.

Here is some background
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/291#issuecomment-461509013

Here are two name proposals
fluent - https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/359
vector - https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/340

> Why are patches occurring in a "vendor" directory?  Is that with the intent to upstream these patches to their source?

Right now the vendor directory is wholly owned by firefox-health. Our frontend-infra discussions will guide what the canonical directory will look like. 
 
> Why are the query strings for this particular application unique and requiring handling of type conversion?  

Since a React app passes query parameters to itself it should have `obj` be structurally equal to `parse(stringify(obj))`.  This identity keeps the code clean because it need not compensate for the lossy conversion.

> The `qs` library parses the "80%" of queries, and using that has the advantage of forcing the app to use query strings in a way that is not surprising to others.  

Excellent! if the qs library works then we can use that. I will see how it passes the test cases (which I will also relax).

> Is there a possibility to simplify the formatting of the query strings wherever they are generated, so that a thin wrapper over `qs` could parse them?

I see no distinction between a "thin wrapper" and a custom library, especially in the face of a test suite:  Both expose a new, unfamiliar, API.  If we can avoid the thin wrapper, then everyone is happy.

> Speaking of surprise, type inference on untrusted data could lead to unexpected behavior, which could lead to security issues.  Obviously not in a frontend-only app, but if this library were used on the backend, then a parameter where a user chooses to write "0" or "null" instead of the expected string might result in a falsy value where one is not expected.

This is only meant for frontend dashboard code; there are no plans to make servers. Even so, verifying query encoding and verifying logical parameters are different levels of abstraction which I imagine are verified in different parts of the code.

> How would a user pass the string "null" using this library?

    http://example.com?a=null

you are right to point out that inconsistency.

> Why not use an existing type-preserving and well-tested encoding like JSON?

I was working under the assumption that the query string be humane. Without being bound so such concerns, the encoding logic is easier:

    {"test":42}  <=>  http://example.com?%7B%22test%22%3A42%7D

PR #359, opened 2019-03-19
rename to Fluent

klahnakoski:
In an effort to find the right name: How about "fluent"?

https://nikas.praninskas.com/javascript/2015/04/26/fluent-javascript/

### DIFF
--- a/src/components/ChartJsWrapper/index.jsx
+++ b/src/components/ChartJsWrapper/index.jsx
@@ -5,7 +5,7 @@ import CircularProgress from '@material-ui/core/CircularProgress';
 import { withStyles } from '@material-ui/core/styles';
 import generateOptions from '../../utils/chartJs/generateOptions';
 import { ErrorMessage } from '../../vendor/errors';
-import { frum } from '../../vendor/queryOps';
+import { fluent } from '../../vendor/fluent';
 
 const styles = {
   // This div helps with canvas size changes
@@ -53,7 +53,7 @@ const ChartJsWrapper = ({
 
     data.datasets.forEach(dataset => {
       const latestDataDate = new Date(
-        frum(dataset.data)
+        fluent(dataset.data)
           .sort('x')
           .last().x
       );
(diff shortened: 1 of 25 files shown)


klahnakoski:
this PR requires merges from other branches.  So let us just discuss the name change, not the code.

armenzg:
If it comes to naming, what @sarah-clements recommends seems more clear.

On the topic of method chaining and fluent interfaces, I've found some blog posts that warn us against using this type of programming:
https://www.yegor256.com/2018/03/13/fluent-interfaces.html
https://ocramius.github.io/blog/fluent-interfaces-are-evil/

I can see how it can read a bit better, however, the maintainabilty cost is higher.

klahnakoski:
Fluent Interfaces break Encapsulation
    Fluent Interfaces break Decorators (and sometimes Composition)
    Fluent Interfaces are harder to Mock
    Fluent Interfaces make diffs harder to read
    Fluent Interfaces are less readable (personal feeling)
    Fluent Interfaces cause BC breaks during early development stages

klahnakoski:
Thank you Armen, I did not see your response when I closed this.  To summarize:

[Fluent Interfaces Are Bad for Maintainability](https://www.yegor256.com/2018/03/13/fluent-interfaces.html) biggest point is that the class used to implement the fluent interface can get big, and I can agree the class does get big. I am not convinced a big class is a maintainability problem: The methods in the class will exists as a method somewhere; whether in a single class or scattered throughout the codebase, the program complexity is no worse.  Plus, this page does not consider [extension methods](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods);  Allowing the fluent class to remain simple, by delegating the method contracts to other parts of the codebase; no different than declaring a vanilla method. 

[Fluent Interfaces are Evil](https://ocramius.github.io/blog/fluent-interfaces-are-evil/) seems to not apply in this case; either it makes assumptions about the definition of fluent, or assumptions about language restrictiveness, or blames poor programming choices on fluent.  Anyway, this page does serve as a good checklist to avoid pathology.

I will change the name to `chainFrom`

PR #361, opened 2019-03-20
Custom tooltip - alternative approach

### LINKED ISSUE #213: Enhanced data when hovering over data points

### LINKED ISSUE #303: Add hg and job link inside tooltip metadata

2alin:
Fixes #303 and originally fixes #213 

@klahnakoski 
As I mentioned in the IRC channel, I'm basically mounting `CustomTooltip` next to `ChartJsWrapper` in the `PerfherderGraphContainer`, so it discards the use of DOM injection as it was handled before (which sends `generateCustomTooltip` to the bin). 

I was forced to use a class component in the `CustomTooltip` because we need to get the dimensions (height and weight) of the component once it's mounted in order to place the tooltip in the right position. And fortunately, this change made the tooltip transitions smooth again.

I'm writing more comments in the code.

Please let me know if this approach is better or worse than the previous one.

Edit: I haven't added the suggestions you made the last time, I'll add them either in this branch or the other one, once you decide which one will be the one to keep working on.


src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -22,19 +23,38 @@ const styles = () => ({
...
   state = {
     data: null,
+    tooltipModel: null,
+    canvas: null,
   };
 
   async componentDidMount() {
-    this.fetchSetData(this.props);
+    this.fetchSetData(this.props).then(() => {
+      const { handleTooltip } = this;
```
2alin: This was the only way that I could think of to call `this.handleTooltip` inside `options.tooltips.custom` without messing with the `this` that's related to the chart when that custom function is called by ChartJS

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -22,19 +23,38 @@ const styles = () => ({
...
+
+      this.setState(prevState => {
+        const { options } = prevState;
+
+        options.tooltips.custom = function custom(tooltipModel) {
+          // eslint-disable-next-line no-underscore-dangle
+          handleTooltip(tooltipModel, this._chart.canvas);
```
2alin: What I'm doing here is binding the environment of the chart (tooltipModel, canvas) with the environment of the `PerfherderGraphContainer` so I can mount/unmount `CustomTooltip` when ChartJS says that it should show up.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,117 @@
...
+    } else if (tooltipModel.yAlign === 'center') {
+      top -= 0.5 * this.state.height;
+    }
+
+    if (tooltipModel.xAlign === 'right') {
+      left -= this.state.width;
+    } else if (tooltipModel.xAlign === 'center') {
+      left -= 0.5 * this.state.width;
+    }
+
```
2alin: I added new cases for vertical and horizontal alignment. Previously (in the other custom tooltip branch), it was only placed at the top or the bottom of the data point. Now it's placed in 3x2 positions all around the data point, which is the exact same implementation as the normal tooltips in ChartJS.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,117 @@
...
+    let top = canvas.offsetTop + tooltipModel.caretY;
+    let left = canvas.offsetLeft + tooltipModel.caretX;
+
+    if (tooltipModel.yAlign === 'bottom') {
+      top -= this.state.height;
+    } else if (tooltipModel.yAlign === 'center') {
+      top -= 0.5 * this.state.height;
+    }
+
+    if (tooltipModel.xAlign === 'right') {
```
klahnakoski: `if/else` blocks should always trigger the question: "why?"

The words "left", "center", and "right" refer to relative positions (0.0, 0.5, 1.0) respectively.  That is not only true here, 
but everywhere we use CSS.  With `CSS = {left: 0, center:0.5, right: 1}`, we can write one line:

    left -= CSS[tooltipModel.xAlign]*this.state.width;

Only make this change if you want.

src/containers/PerfherderGraphContainer/index.jsx
2alin: All of this, checked. Except this line
```
const { canvas } =  self._chart;
```
The context of `self` is not inside the custom function so there's no `self._chart`.  I have another idea to avoid using `this._chart.canvas`. I'll see if I can implement it in the next pack of commits or discuss it with you first.

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -52,6 +72,13 @@ class PerfherderGraphContainer extends Component {
           options={options}
           missingDataError
         />
+        {tooltipModel && (
+          <CustomTooltip
+            tooltipModel={tooltipModel}
+            series={options.series}
+            canvas={canvas}
```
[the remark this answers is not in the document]
2alin: Yup, we need `canvas` in order to get the position of the element (canvas) where the chart is being rendered and then calculate the position of the tooltip. `tooltipModel` only gives the relative position of the datapoint with respect to the canvas. And the only way to get it in a direct way is through `this._chart`. I already looked through `series`, `data`, and `tooltipModel`and there's no way to get it from there.

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -52,6 +72,13 @@ class PerfherderGraphContainer extends Component {
           options={options}
```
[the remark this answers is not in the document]
2alin: The reason I defined `options.tooltips.custom` inside `PerfherderGraphContainer` is so I could have access to `tooltipsModel`because it's the only way we can know if the tooltip is going to show up or not. If you declare such custom function anywhere else, you'll need to link to the argument of the function `tooltipsModel` and bring it back to `PerfherderGraphContainer` somehow.

There's something else. Until now we were only modifying `perfherderFormatter` which is like `generateOptions` but specific to perfherder (last two charts at the bottom). If we modify `generateOptions` we will change the behaviour in all the charts not just the perfherder. 

In my opinion, `CustomTooltip` should be inside `ChartJsWrapper`, next to `<Chart />`, but because the original issue was only refered to the perfherder, I thought that solving this issue would be an experiment, and once it's fully implemented, we could move the code to `ChartJsWrapper`and apply it anywhere else.

src/utils/chartJs/CustomTooltip.jsx
2alin: Nice one, I love how there's always something to improve :smiley: 
I will change only the name ('CSS'), because we also need  another for 'top', 'center', 'bottom'. But before that what about using this:
```
let top = canvas.offsetTop + tooltipModel.caretY;
top -=
      { top: 0, center: 0.5, bottom: 1 }[tooltipModel.yAlign] *
      this.state.height;
```
or even more compact:
```
const top =
      canvas.offsetLeft + tooltipModel.caretX -
      { top: 0, center: 0.5, bottom: 1 }[tooltipModel.yAlign] *
        this.state.height;
```
would be alright? or is it too messy?

2alin:
@klahnakoski 
Now, for the alternative way to using `this._chart.canvas` and passing it to `CustomTooltip`as props, we can get such canvas element inside `CustomTooltip` traversing the DOM tree and using the reference we had set already there (`this.tooltip`), with this:
```
this.tooltip.current.parentElement.getElementsByTagName('canvas')
```
what do you think? is it better? I think it is.

src/utils/chartJs/CustomTooltip.jsx
klahnakoski: yes, either is fine.  I like that you inlined the "CSS" structure.

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -22,10 +23,25 @@ const styles = () => ({
...
+    await this.fetchSetData(this.props);
+
+    const self = this;
+
+    this.setState(prevState => {
+      const { options } = prevState;
+
+      options.tooltips.custom = function custom(tooltipModel) {
+        // eslint-disable-next-line no-underscore-dangle
+        self.setState({ tooltipModel, canvas: this._chart.canvas });
```
klahnakoski: One more thing, now that this is inlined, I can see that we only need to set ` canvas: this._chart.canvas` once.  Can it be done outside of `custom()`, just above?

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -0,0 +1,116 @@
...
+    width: '10px',
+    height: '10px',
+    marginRight: '10px',
+  },
+};
+
+class CustomTooltip extends React.Component {
+  constructor(props) {
+    super(props);
+    this.tooltip = React.createRef();
```
klahnakoski: I do not understand why you need this. I would assume this component will render to whatever size the `render()` produces.

klahnakoski:
@2alin also, add ?some CSS? so the tooltip is not so close to the datapoint

image

src/containers/PerfherderGraphContainer/index.jsx
2alin: It's not possible. Again, because this._chart only exists inside the custom function. 

The custom function is not being executed right away, it will be executed when the tooltip pops up by ChartJS, and what we are doing is creating the custom function here so we can bind those two properties (`this._chart`, `tooltipModel`) to the `PerfherderGraphContainer` and do magic as you established before.

I know it doesn't look good, but for now it's the cleanest way I can imagine to bind chartjs behaviour with our react components. I'm still not giving up, and if I come with something else in the future, I'll let you know :)

2alin: Also, I like your idea about using HOC. I started yesterday already reading about them and yup I'll try to use them in the next PR.

src/utils/chartJs/CustomTooltip.jsx
2alin: I suppose you are asking about `this.tooltip = React.createRef()`. We need a way to measure the width and height of the CustomTooltip in order to locate it in the 3x2 positions around the data point (e.g. if you want to move the tooltip over the data point, first get the position of the chart, then add position of the data point, then subtract height of tooltip). 
And because we are styling such location inside the same component, I only thought of React Refs to target and access such DOM information.

2alin:
> @2alin also, add ?some CSS? so the tooltip is not so close to the datapoint
> 
> image

Yup, that's part of my work here for today. I'll try to replicate the caret triangle of the original tooltip :grin:

src/utils/chartJs/CustomTooltip.jsx
klahnakoski: ok, this is good enough for now.  I would imagine that we can mess with the CSS absolute positioning (relative to the point) instead of doing actual math (like is done now).

klahnakoski:
If you have nothing else, then I will merge this.  This tooltip is better than what we have now, and any improvements can be be done in other PRs.  

There is merge conflict, if you have not solved this by tomorrow, then I will. 

Thank you very much for the work you have put into this!!

src/utils/perfherder/chartJs/perfherderFormatter.js
```diff
@@ -9,8 +9,12 @@ const dataToChartJSformat = data =>
     y: value,
   }));
 const generateInitialOptions = series => {
-  // The first series' metadata defines the whole set
-  const higherIsBetter = !series[0].meta.lower_is_better;
+  // TODO: map tests and suite scores to measurement units and
+  // add some label for scale
+  const isTest = !missing(series.meta.test);
```
2alin: it should be now
```
const isTest = !missing(series[0].meta.test);
```

2alin:
Welcome @klahnakoski. 
There's a small change that was left to do, and I marked in the code review.

After that, it looks like now in the current state of the repo, the `options.tooltips.custom` is not being rendered and then tooltip doesn't show up. I've tracked down some code, and `ChartJsWrapper` is receiving `options.tooltips.custom`and send it to chartjs as options. But still not being called.

If you can help me with that, I'll be very grateful. Otherwise I'll try to fix it in the days to come.

2alin:
@klahnakoski
Couldn't rest without fixing it :sweat_smile: 
I think now it's ready to ship.

klahnakoski:
checking one last time.... starting now.

klahnakoski:
Thanks for fixing the merge conflicts.  Sorry I did not look at this yesterday.

It seems there are still some not-immediate problems: Pages will go blank.  But I think this page is consistently failing:

http://localhost:5000/android/tp6m?test=loadtime&platform=android-hw-g5-7-0-arm7-api-16

image

You may have noticed that the tooltip is a bit indecisive about if will show.  This is not a big problem, but there is also the issue of the color of 

    <span style={labelColors} className={classes.tooltipKey} />

which is missing, sometimes:

image

they may be related.

2alin:
Didn't notice the errors about the pages going black. I'll check it.



And about this:

> but there is also the issue of the color of
> 
> ```
> <span style={labelColors} className={classes.tooltipKey} />
> ```
> 
> which is missing, sometimes:

That's a general behaviour in all charts in this repo. When the cursor is exactly over the data point (not just close), the (background) fill of such square becomes white. Look at this normal chart (square fill is white, though border is yellow):

Screenshot_20190326_161651

I noticed, in the past, the exact code line(s) that was producing this behaviour, and I found it meaningless, but I leave it like that. I'll try to find those lines again and fix it.

2alin:
@klahnakoski 
And about this:
> You may have noticed that the tooltip is a bit indecisive about if will show. This is not a big problem,

Yup I noticed it, but that happens in all charts (independently of tooltip) where the dataset is not big enough to make it looks like there's no horizontal gaps between each data point. Look at the next chart (no perfhender), same happens, cursor is close to make the data point active but tooltip doesn't show up (even if it's not a custom tooltip). I'll see if something can be done about that, but no promises.

Screenshot_20190326_165125

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -23,25 +24,45 @@ const styles = () => ({
...
       this.setState({ isLoading: true });
       this.setState(await getPerfherderData(series));
     } finally {
-      this.setState({ isLoading: false });
+      const self = this;
+
+      this.setState(prevState => {
+        const { options } = prevState;
+
+        options.tooltips.custom = function custom(tooltipModel) {
```
klahnakoski: The `options` is undefined, and this method throws an error.  Please add a check for that case.

klahnakoski: https://deploy-preview-361--firefox-health-dashboard.netlify.com/android/tp6m?test=loadtime&platform=android-hw-g5-7-0-arm7-api-16

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -48,8 +48,10 @@ class PerfherderGraphContainer extends Component {
           self.setState({ tooltipModel, canvas: this._chart.canvas });
         };
 
-        return { ...prevState, options, isLoading: false };
+        return { ...prevState, options };
       });
+    } finally {
+      this.setState({ isLoading: false });
     }
   }
```
2alin: My mistake, setting `options.tooltips.custom`  should be done only if `getPerfherderData(series)` was successful, not in `finally{}`. :sweat_smile:

klahnakoski:
Excellent work!

2alin:
Awesome!! Thanks for all the reviews and the merge. :grinning:

aimenbatool:
@2alin congratulations. :-) Good work (y) 

I'm wondering why am I unable to click and lock the tooltip? I am unable to lock the tooltip and click the link to check if it is working or not.

@klahnakoski, @armenzg did requirements change?

armenzg:
I'm having the same problem. If we can't fix it easily I suggest we back out the code until then.

2alin:
@aimenbatool @armenzg @klahnakoski 
Yup, you are right, the locking feature wasn't implemented yet, as well as some other features of the custom tooltip.I think the partial work was merged because it was already very long. 

I was planning to dedicate some hours today to finish the other features first and then the locking one. But because this is important, I'll dedicate the whole day to have everything ready today, even the locking feature.

PR #373, opened 2019-03-30
Add new features to custom tooltip

### LINKED ISSUE #370: Fix tooltip's wrong placement
The custom tooltip is not updating its state's width/height, used to position itself, after the first time that shows up (mounts), which give us an inappropriate placement for some data points:

Screenshot_20190328_200254

### LINKED ISSUE #371: Add arrow tip to tooltip
The tip should change orientation and placement, depending on the tooltip position (8 different ones).

### LINKED ISSUE #372: click-lock is not working on new tooltip
as per

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/361#issuecomment-477944311

some code involving clicking can be found here:

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/utils/registerTooltip.jsx

2alin:
This fixes #370, fixes #371, fixes #372 and continues #361. 

@klahnakoski 
It looks quite amazing, I think you will feel as proud as I feel right now. :grin: 

Tooltip on action (12s): https://youtu.be/2j-5VvuG08Q

__To consider__: sometimes the tooltip arrow looks like it's in the wrong position (a little) compared to the data point, but that's chartjs fault. Just for testing, I rendered the custom and default tooltips at the same time, and both arrow tips match (almost) perfectly: (39s) https://youtu.be/Vn5xAwOEHWs (recorded before the locking feature).

More comments in the code.

ping: @armenzg @aimenbatool

### DIFF
--- a/src/containers/PerfherderGraphContainer/index.jsx
+++ b/src/containers/PerfherderGraphContainer/index.jsx
@@ -25,6 +25,7 @@ class PerfherderGraphContainer extends Component {
   state = {
     data: null,
     tooltipModel: null,
+    tooltipIsLocked: false,
     canvas: null,
     isLoading: false,
   };
@@ -43,9 +44,12 @@ class PerfherderGraphContainer extends Component {
       this.setState(prevState => {
         const { options } = prevState;
 
+        options.onClick = this.handleChartClick;
         options.tooltips.custom = function custom(tooltipModel) {
-          // eslint-disable-next-line no-underscore-dangle
-          self.setState({ tooltipModel, canvas: this._chart.canvas });
+          if (!self.state.tooltipIsLocked) {
+            // eslint-disable-next-line no-underscore-dangle
+            self.setState({ tooltipModel, canvas: this._chart.canvas });
+          }
         };
 
         return { ...prevState, options };
@@ -54,6 +58,12 @@ class PerfherderGraphContainer extends Component {
       this.setState({ isLoading: false });
     }
   }
+
+  handleChartClick = () => {
+    this.setState(prevState => ({
+      tooltipIsLocked: !prevState.tooltipIsLocked,
+    }));
+  };
 
   render() {
     const { classes, title } = this.props;
@@ -64,6 +74,7 @@ class PerfherderGraphContainer extends Component {
       canvas,
       tooltipModel,
       isLoading,
+      tooltipIsLocked,
     } = this.state;
 
     return (
@@ -87,6 +98,7 @@ class PerfherderGraphContainer extends Component {
             tooltipModel={tooltipModel}
             series={options.series}
             canvas={canvas}
+            isLocked={tooltipIsLocked}
           />
         )}
       </div>
--- a/src/utils/chartJs/CustomTooltip.jsx
+++ b/src/utils/chartJs/CustomTooltip.jsx
@@ -12,6 +12,9 @@ const styles = {
     borderRadius: '4px',
     pointerEvents: 'none',
     transform: 'translate(var(--trans-x), var(--trans-y))',
+    '& a': {
+      pointerEvents: 'auto',
+    },
     '&::before': {
       position: 'absolute',
       content: '""',
@@ -70,11 +73,14 @@ const styles = {
     height: '10px',
     marginRight: '10px',
   },
+  lockMessage: {
+    color: '#ccc',
+  },
 };
 
 class CustomTooltip extends React.Component {
   render() {
-    const { classes, tooltipModel, series, canvas } = this.props;
+    const { classes, tooltipModel, series, canvas, isLocked } = this.props;
 
     if (tooltipModel.opacity === 0) return null;
 
@@ -141,6 +147,9 @@ class CustomTooltip extends React.Component {
             job
           </a>
           )
+        </div>
+        <div className={classes.lockMessage}>
+          {isLocked ? 'Click to unlock' : 'Click to lock'}
         </div>
       </div>
     );
--- a/src/utils/chartJs/CustomTooltip.jsx
+++ b/src/utils/chartJs/CustomTooltip.jsx
@@ -3,12 +3,68 @@ import { withStyles } from '@material-ui/core/styles';
 
 const styles = {
   tooltip: {
+    '--bg-color': 'rgba(100, 0, 0, .8)',
+    '--tip-size': '6px',
     position: 'absolute',
     padding: '6px',
-    background: 'rgba(0, 0, 0, .8)',
+    background: 'var(--bg-color)',
     color: 'white',
     borderRadius: '4px',
     pointerEvents: 'none',
+    transform: 'translate(var(--trans-x), var(--trans-y))',
+    '&::before': {
+      position: 'absolute',
+      content: '""',
+      borderWidth: 'var(--tip-size)',
+      borderStyle: 'solid',
+      borderColor: 'transparent',
+    },
+
+    '&.ytop': {
+      color: 'blue',
+      '--trans-y': 'var(--tip-size)',
+      '&::before': { bottom: '100%', borderBottomColor: 'var(--bg-color)' },
+      '&.xright': {
+        '--trans-x': 'calc(-100% + 2 * var(--tip-size))',
+        '&::before': { right: 'var(--tip-size)' },
+      },
+      '&.xcenter': {
+        '--trans-x': '-50%',
+        '&::before': { left: 'calc(50% - var(--tip-size))' },
+      },
+      '&.xleft': {
+        '--trans-x': 'calc(-2 * var(--tip-size))',
+        '&::before': { left: 'var(--tip-size)' },
+      },
+    },
+    '&.ycenter': {
+      '--trans-y': '-50%',
+      '&::before': { top: 'calc(50% - var(--tip-size))' },
+      '&.xright': {
+        '--trans-x': 'calc(-1 * (100% + var(--tip-size)))',
+        '&::before': { left: '100%', borderLeftColor: 'var(--bg-color)' },
+      },
+      '&.xleft': {
+        '--trans-x': 'var(--tip-size)',
+        '&::before': { right: '100%', borderRightColor: 'var(--bg-color)' },
+      },
+    },
+    '&.ybottom': {
+      '--trans-y': 'calc(-100% - var(--tip-size))',
+      '&::before': { top: '100%', borderTopColor: 'var(--bg-color)' },
+      '&.xright': {
+        '--trans-x': 'calc(-100% + 2 * var(--tip-size))',
+        '&::before': { right: 'var(--tip-size)' },
+      },
+      '&.xcenter': {
+        '--trans-x': '-50%',
+        '&::before': { left: 'calc(50% - var(--tip-size))' },
+      },
+      '&.xleft': {
+        '--trans-x': 'calc(-2 * var(--tip-size))',
+        '&::before': { left: 'var(--tip-size)' },
+      },
+    },
   },
   tooltipKey: {
     display: 'inline-block',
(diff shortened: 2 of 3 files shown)


src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -87,6 +98,7 @@ class PerfherderGraphContainer extends Component {
             tooltipModel={tooltipModel}
             series={options.series}
             canvas={canvas}
+            isLocked={tooltipIsLocked}
```
2alin: I'm sending `isLocked` as props to the tooltip in order to throw a customized message to the user.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -1,57 +1,85 @@
...
+  },
+  '&.xcenter': {
+    '--trans-x': '-50%',
+    '&::before': { left: 'calc(50% - var(--tip-size))' },
+  },
+  '&.xleft': {
+    '--trans-x': 'calc(-2 * var(--tip-size))',
+    '&::before': { left: 'var(--tip-size)' },
+  },
+};
```
2alin: For 'top' and 'bottom' alignment, the custom position style share a lot of similarities (except two lines). The only way I could think of to share the same code in both cases was to define one of them as constant and share it.

Check (if you wish) my [raw diagrams](https://docs.google.com/presentation/d/1i-Xz7z6NqzQwiWYsgtjvrZP-t6JAbzfMfA1K0b1eG7w/edit?usp=sharing) to see the alignments.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -1,57 +1,85 @@
...
 const styles = {
   tooltip: {
+    '--bg-color': 'rgba(0, 0, 0, .8)',
+    '--tip-size': '6px',
```
2alin: This is awesome, just change the value of `--tip-size` and all the adjustments needed, for that size, are made automatically.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -1,57 +1,85 @@
...
     position: 'absolute',
     padding: '6px',
-    background: 'rgba(0, 0, 0, .8)',
+    background: 'var(--bg-color)',
     color: 'white',
     borderRadius: '4px',
     pointerEvents: 'none',
+    transform: 'translate(var(--trans-x), var(--trans-y))',
+    '& a': {
+      pointerEvents: 'auto',
```
2alin: We need to exclude anchors from our previous `pointerEvents: 'none'`, otherwise they wouldn't be clickable. And we need the last one in order to hover datapoints under the tooltip.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -1,57 +1,85 @@
...
+        '--trans-x': 'var(--tip-size)',
+        '&::before': { right: '100%', borderRightColor: 'var(--bg-color)' },
+      },
+    },
+    '&.ybottom': {
+      ...topAligned,
+      '--trans-y': 'calc(-100% - var(--tip-size))',
+      '&::before': { top: '100%', borderTopColor: 'var(--bg-color)' },
+    },
   },
```
2alin: You probably noticed that I used 'ytop, ycenter, ytop, xleft, xcenter, xright'. Why not just 'top, center, etc'? Mainly because we have two 'center' for horizontal and vertical position, and they could overwrite each other. 

And having 'top, ycenter, bottom' could give us a little of trouble with how we generate the class names at a lower part of the code. Arguable decision.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -1,57 +1,85 @@
...
-    this.state = { width: 0, height: 0 };
-  }
-
-  componentDidMount() {
-    this.setState({
-      width: this.tooltip.current.clientWidth,
-      height: this.tooltip.current.clientHeight,
-    });
-  }
-
```
2alin: I think you'll be happy that we were able to get rid of this :sweat_smile:

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -110,6 +140,9 @@ class CustomTooltip extends React.Component {
           </a>
           )
         </div>
+        <div className={classes.lockMessage}>
+          {isLocked ? 'Click to unlock' : 'Click to lock'}
+        </div>
```
2alin: As the [treeherder graphs](https://treeherder.mozilla.org/perf.html#/graphs?series=mozilla-central,1776360,1,10&series=mozilla-central,1775979,1,10&series=mozilla-central,1776166,1,10&selected=mozilla-central,1776166,413555,667080523,10), I think it's nice to provide the user with feedback of how to lock the tooltips. I gave it a gray color in the style to make it less intrusive, and if you still find it too much we can discuss about it.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -1,57 +1,85 @@
...
+    '--bg-color': 'rgba(0, 0, 0, .8)',
+    '--tip-size': '6px',
     position: 'absolute',
     padding: '6px',
-    background: 'rgba(0, 0, 0, .8)',
+    background: 'var(--bg-color)',
     color: 'white',
     borderRadius: '4px',
     pointerEvents: 'none',
+    transform: 'translate(var(--trans-x), var(--trans-y))',
```
klahnakoski: **this** is blows my mind!  Let me make sure I get this straight (and only for `x` since `y` is similar)

* the `tooltipModel.xAlign` is used to choose the class (let me call it the "xClass").
* the choice of xClass sets the `--trans-x` variable value
* the `transform` uses `--trans-x` to position the tooltip

Can `tooltipModel.xAlign` set the `transform` x-coordinate directly?  I understand that would mean we still have math in the code, but piping coordinate calculations through class names is convoluted.  

I will say, what you have done is an impressive display of CSS skills. My CSS skills are poor, so you have taught me a few things.  Thank you.  

In general, I appreciate the work you have done in getting the element to position properly, for this reason I am inclined to accept this PR now because it works rather than invest more hours into "simplifying" the code. Even with "simplification" you will still be fighting with the HTML/CSS box model, and the complicated math that it demands, and code that switches based on property names.  Really, not much better.  

There is a better way, [which I have written about](https://github.com/klahnakoski/jsLayout). My code does not work in a JSX React environment (because it manipulates HTML elements directly), but the idea should be translated in to a library that will allow for easier alignment expressions.

For example, Let `p`, `a`, and `t` represent the point, arrow and tooltip respectively:

for top-right positioning 

image

we have 

    p.mc = a.mc
    a.tl = t.bl

more explicitly 

    p.middle = a.middle
    p.center = a.center
    a.top = t.bottom
    a.left = t.left

this ignores the `cornerspace`, but that can be solved with more `divs`, or using `box-sizing` to clarify the alignment.  

That is all a dream for the future, but I hope you keep it in mind when you are positioning elements.

klahnakoski:
@aimenbatool may you check to see if anything else is missed? I am unable to set you as a reviewer formally. 

thank you!

src/utils/chartJs/CustomTooltip.jsx
2alin: you are right, I kept thinking that I was dealing only with scss.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -1,57 +1,85 @@
...
-    }
-
-    if (tooltipModel.xAlign === 'right') {
-      left -= this.state.width;
-    } else if (tooltipModel.xAlign === 'center') {
-      left -= 0.5 * this.state.width;
-    }
-
+    const top = canvas.offsetTop + tooltipModel.caretY;
+    const left = canvas.offsetLeft + tooltipModel.caretX;
```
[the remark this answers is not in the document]
2alin: I love your challenges.
I've previously encountered [JSS Dynamic Values](https://cssinjs.org/react-jss?v=v10.0.0-alpha.16#dynamic-values) and after reading your comment, I tried to applied it to our code so we could get rid of not just calculations but all the inline style made inside the component. Unfortunately I wasn't lucky to make it work at the moment, not even with testing commands like `color: props => 'blue'`. I'll see what am I doing it wrong, and if I can't make it work in this PR, it will be in the next ones.

src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -1,57 +1,85 @@
 import React from 'react';
 import { withStyles } from '@material-ui/core/styles';
 
+const topAligned = {
```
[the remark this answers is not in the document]
2alin: I don't think so, `xAligned`makes me think of different horizontal positions but independently of the vertical alignment. And the style in `topAligned` is exactly the one needed for the 'top' vertical alignment. Also the style for the three horizontal alignments here doesn't apply to the 'center' vertical one.

src/utils/chartJs/CustomTooltip.jsx
2alin: > * the `tooltipModel.xAlign` is used to choose the class (let me call it the "xClass").
> * the choice of xClass sets the `--trans-x` variable value
> * the `transform` uses `--trans-x` to position the tooltip

Yup, that's exactly what it does, as well as to adjust the triangle tip (`::before`).

> Can `tooltipModel.xAlign` set the `transform` x-coordinate directly? I understand that would mean we still have math in the code, but piping coordinate calculations through class names is convoluted.

You do love perfecting stuff :sweat_smile: . Yup yup, as I mentioned in a previous comment [JSS Dynamic values](https://cssinjs.org/react-jss?v=v10.0.0-alpha.16#dynamic-values) would help us move all the style related stuff outside the component, but I wasn't able to make it work in my first try, I'll keep working on it in my spare time, don't worry.

>  My CSS skills are poor, so you have taught me a few things. Thank you.

Thank YOU! As I have mentioned many times before, all your previous reviews, your comments, and the way you look for perfection is what makes me learn more and search for new ways to improve what's been done, and I freaking enjoy that.

Checking your method...

2alin: Wow just wow. I just checked your jsLayout library, and I'm surprised that it doesn't have the popularity that should deserve. It gives you a 'ping' and 'elastic' feeling. Like a canvas api with steroids, for DOM elements. Love it.


> That is all a dream for the future, but I hope you keep it in mind when you are positioning elements.

Yup yup, It's already stuck in my mind and I'll see where and how your idea can be applied in this project or others. Thank you!!

src/utils/chartJs/CustomTooltip.jsx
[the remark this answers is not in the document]
2alin: @klahnakoski 
I finally made it work, *but* we need to install `@material-ui/styles`, and import `withStyles` from `@material-ui/styles` instead of `@material-ui/core/styles`.

As it's stablished [here](https://material-ui.com/css-in-js/basics/) (at the top in the warning note), `@material-ui/styles` is still an alpha version, so I'm not sure if we should add that dependency now. Let me know what should be best, please.

Sources: [original issue](https://github.com/mui-org/material-ui/issues/7633) and [merged PR](https://github.com/mui-org/material-ui/pull/13503).

src/utils/chartJs/CustomTooltip.jsx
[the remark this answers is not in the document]
klahnakoski: name does not matter

src/utils/chartJs/CustomTooltip.jsx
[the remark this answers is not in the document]
klahnakoski: My mistake.  I assumed the`@material-ui/core/styles` was capable of accepting dynamic values.

* Can **we** simply set the HTML element `style` property without using classes, and without using any of these CSS tools?
* What is the compiled size difference when using `@material-ui/styles` vs `@material-ui/core/styles` , or any other dynamic CSS solution?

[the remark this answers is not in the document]
klahnakoski: save it for the next PR

klahnakoski:
@2alin  We will wait for @aimenbatool to chime in.  If you have any other PRs, be sure to base them off of this branch, which should be merged by Monday morning.

src/utils/chartJs/CustomTooltip.jsx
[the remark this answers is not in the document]
2alin: Got it,  I'll work on it for the next PR. But answering your questions, anyway:

> Can **we** simply set the HTML element `style` property without using classes, and without using any of these CSS tools?

Already thinking about this.

> What is the compiled size difference when using `@material-ui/styles` vs `@material-ui/core/styles` , or any other dynamic CSS solution?

`./build` (using `@material-ui/styles`and without it): **8.7M** vs **8.4M**

aimenbatool:
@2alin @klahnakoski Great work. This looks perfect. 

I will take some time out and definitely read the code. 👏👏👏 

A quick thing: When we click one data point and lock it then we can't see any other data point on mouse hover. 
Here is what is happening:
- We can see data details on first time hovering
- We click & lock data  point
- We now move the mouse and this time we don't see data details on mouse hover (which was happening earlier)
- We can see data details upon clicking only after first click

Drawback: We will have to click an unknown data point and lock it to see details. 

[This is the behavior](https://deploy-preview-242--firefox-health-dashboard.netlify.com/android) I'm talking about. Deploy https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/242 and check.

@armenzg  can tell better if it is good to add it back or not.

Rest is an awesome work @2alin :)

klahnakoski:
@aimenbatool thank you for noticing that subtle behaviour.  I added it as another issue.

PR #383, opened 2019-04-01
use scatter for perfherder charts, keep others as line

klahnakoski:


klahnakoski:
Will try again in another pr

PR #384, opened 2019-04-02
Fix crash because of empty slots for series array

### LINKED ISSUE #382: Sometimes currSeries does not exist
This line will crash the page because `currSeries` is not defined

    const higherOrLower = currSeries.meta.lower_is_better

We can add a check, but the real problem is the array is not filled.

2alin:
**Update:** This was an attempt to patch #382, but, as Kyle clearly mentioned:
> By filtering our missing series you change the index of each series, which then changes the color of the series.

it could give us more trouble, so the PR has been rejected.

----------------

@klahnakoski 
As you well mentioned in your issue, the real problem was getting empty slots in the series array. And the reason of that is the way [`getPerherderData`](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/utils/perfherder/chartJs/getPerfherderData.js) creates `newData`(which will become `series` once it gets formatted by `perfherderFormatter`): 

First it creates an array with empty slots taking the length of a primitive `series` ([L7](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/utils/perfherder/chartJs/getPerfherderData.js#L7)). Then it fills each slot with data obtained from `queryPerformanceData` ([L17](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/utils/perfherder/chartJs/getPerfherderData.js#L17)), but the problem is here, because `queryPerfomanceData` doesn't return data for all the series in all charts (_check [1] to see the series/charts affected and check [2] to read my deeper tracking analysis_).

Now I think that independently of `queryPerformanceData` returning data or not, `getPerherderData` **shouldn't return an array with empty slots**. That's the reason I decided to apply a fix here, which is removing such empty slots before returning `newData`(remember, it becomes `series`after being formatted). 

I also thought on changing the way `newData` is built, creating an empty array first, and pushing elements in it while being 'fetched', but because such creation is asynchronous, we would lose the original order of each series, which is visually important in charts labels. I'm a newbie with asynchronous functions but I think that would happen and it's the reason @armenzg built it that way.

**[1]** 

The charts (suites) being affected are:
"raptor-tp6m-google-geckoview", series Index 1 (one of two series are displayed)
"raptor-tp6m-wikipedia-geckoview", series Index 1 (one of two series are displayed)
"raptor-tp6m-youtube-geckoview", series Index 1 (one of two series are displayed)
"raptor-unity-webgl-geckoview", series Index 3 (four of five series are displayed)

**[2]** 

When I noticed that `queryPerformanceData` ([L17](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/utils/perfherder/chartJs/getPerfherderData.js#L17)) wasn't returning data from the cases in [1]. I tried to track down the problem. This is what I got (only affected series):

-\>  In `perf-goggles.js`, for some `seriesConfig`, because `parentSignatureInfo` produces an empty `parentInfo`, `queryPerfomanceData` returns an empty object ([L232](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/vendor/perf-goggles.js#L232))

-\>  In `parentSignatureInfo`, `signaturesForPlatformSuite(seriesConfig)` ([L201](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/vendor/perf-goggles.js#L201)) is empty.

-\>  In [`signaturesForPlatformSuite`](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/vendor/perf-goggles.js#L142), there's no signature in `allPlatformSignature` with `suite` property that matches `seriesConfig.suite`. For example you could analyze the case for the suite 'raptor-unity-webgl-geckoview' with:
```
const allPlatformSignatures = ...
+ if (seriesConfig.suite === 'raptor-unity-webgl-geckoview') {
+   console.log(allPlatformSignatures);
+ }
...
+console.log(filteredSignatures);
return filteredSignatures;
```
and see that one of the series (index 3) gets missed because [L153](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/vendor/perf-goggles.js#L153) was never satisfied.

And here is where I stopped, because I was already mentally tired (too late in the night).

### DIFF
--- a/src/utils/perfherder/chartJs/getPerfherderData.js
+++ b/src/utils/perfherder/chartJs/getPerfherderData.js
@@ -41,7 +41,7 @@ const getPerfherderData = async series => {
       `can not get data for ${JSON.stringify(series[0].seriesConfig)}`
     );
 
-  return perfherderFormatter(newData);
+  return perfherderFormatter(newData.filter(Boolean));
 };
 
 export default getPerfherderData;


src/utils/perfherder/chartJs/getPerfherderData.js
```diff
@@ -41,7 +41,7 @@ const getPerfherderData = async series => {
       `can not get data for ${JSON.stringify(series[0].seriesConfig)}`
     );
 
-  return perfherderFormatter(newData);
+  return perfherderFormatter(newData.filter(Boolean));
```
2alin: If you approve this fix, I wasn't sure if `newData` should be filtered and saved first:
```
newData = newData.filter(Boolean);
return perfherderFormatter(newData);
```

or what I did.

klahnakoski:
@2alin  It is good that you are finding the root cause, and thank you for for putting time into this. I added a PR to patch over the problem for now, since this your solution must go deeper.

The abstraction used in these dashboards is a bit off: The downloading of data, and the formatting of that data is weaved together tightly.  By filtering our missing series you change the index of each series, which then changes the color of the series.  I would like to decouple formatting from downloading, but that is a big job.

Another problem is the charting components can not handle `nulls` gracefully yet.  Going forward, I would like to send big grids of numbers to the charting library for display. The grids may be sparse; where whole columns or whole rows are null (representing days or series respectively).  These grids can be joined with formatting arrays to provide formatting; each row in a data series matches a row in the formatting array, like we do now but as a vector operation rather than using a loop and index.

Although the charting lib must be robust to nulls, we must still do the work you started to be sure those nulls represent legitimate lack of data rather than some error state.

src/utils/perfherder/chartJs/getPerfherderData.js
klahnakoski: What you did was elegant enough, but it messed with the colors, so we will not use it.

2alin:
@klahnakoski 
> I added a PR to patch over the problem for now, since this your solution must go deeper.

Yup, I already checked your PR, and it fixes the problem from the tooltip side without giving more trouble, which is pretty nice.

> By filtering our missing series you change the index of each series, which then changes the color of the series.

I missed this!! thank you for noticing :sweat_smile:.

>  I would like to decouple formatting from downloading, but that is a big job.

This is interesting, I'll think on it.

> Another problem is the charting components can not handle nulls gracefully yet. Going forward, I would like to send big grids of numbers to the charting library for display. The grids may be sparse; where whole columns or whole rows are null (representing days or series respectively). These grids can be joined with formatting arrays to provide formatting; each row in a data series matches a row in the formatting array, like we do now but as a vector operation rather than using a loop and index.

I have a lot yet to learn, but I freaking love it.

src/utils/perfherder/chartJs/getPerfherderData.js
2alin: Yup, I completely agree, thank you for noticing the issue that would bring to us.

2alin:
And because of the colors issue,  I proceed to close this PR and change the 'fixes' closing issue keyword.
Thank you @klahnakoski.

PR #385, opened 2019-04-02
patch for tooltip crash

klahnakoski:
This is a temporary fix until the root cause, and a solution id found


src/utils/chartJs/CustomTooltip.jsx
```diff
@@ -94,6 +95,12 @@ class CustomTooltip extends React.Component {
     };
     const { index } = currPoint;
     const currSeries = series[currPoint.datasetIndex];
+
+    if (missing(currSeries)) {
+      // TODO: Add a warning here
+      return null;
```
klahnakoski: show no tooltip of the series is missing

Issue #314, opened 2019-02-21
browser back and forward do not update page


klahnakoski:
Prthessing the forward and back buttons on the browser updates he URL, but does not update the page. 

`withNavigation()` should handle this problem

exturingmachina:
Hello, would it be okay if I worked on this?

djmitche:
Go ahead @exturingmachina !

shashankgaurav17:
Hey, Can I work on this ?

klahnakoski:
@shashankgaurav17   yes please!

TrangEmily:
Hi @klahnakoski I can't reproduce this issue. When I pressed backward and forward button on Firefox it reload the page. Is it solved already? Thank you.

klahnakoski:
go to the tp6mobile page:

https://health.graphics/android/tp6m?test=loadtime&platform=android-hw-g5-7-0-arm7-api-16

update the platform to P2, notice reload

press back button on browser, notice the URL changes, but data does not

TrangEmily:
Yes when I go to that pape, forward and backward changes the URL but not reload the page. However, I can not see the data. Is it another issue or I lack something on my installation?
This is on Android page (other parts show normal graph and data)
<img width="1663" alt="Screen Shot 2019-03-29 at 11 20 03 AM" src="https://user-images.githubusercontent.com/42683686/55254430-0dc02d80-5215-11e9-9f9b-097fbf193171.png">
And this is when I click "more" link which go to the link you gave in above comment:
<img width="1663" alt="Screen Shot 2019-03-29 at 11 20 29 AM" src="https://user-images.githubusercontent.com/42683686/55254471-26c8de80-5215-11e9-91f1-1860955fe9bf.png">

klahnakoski:
It looks like you have not network connection.  What does the network tab in the browser say?

TrangEmily:
No I run on local host, other parts show normally as below
<img width="1654" alt="Screen Shot 2019-03-29 at 12 58 56 PM" src="https://user-images.githubusercontent.com/42683686/55259315-8ed1f180-5222-11e9-8fa8-b611bc8bd3b7.png">

klahnakoski:
@TrangEmily  And what does the debugger say?

![image](https://user-images.githubusercontent.com/2334429/55259610-8a1a3700-523c-11e9-8abf-c1176636ee32.png)

TrangEmily:
And this is the network tab
[...]

PR #393, opened 2019-04-03
Solve the backward button error

TrangEmily:
This PR solves issue #314 
Add `// eslint-disable-next-line react/no-did-update-set-state` because without it, ESLint returns this error when commit:
`$ eslint --cache --ext mjs,jsx,js --format codeframe ".*.js" src test /workspace/firefox-health-dashboard/src/vendor/utils/navigation/index.jsx
			Do not use setState in componentDidUpdate (react/no-did-update-set-state) at src/vendor/utils/navigation/index.jsx:59:11:
			57 |           // Update state based on current props
			58 |
			> 59 |           this.setState(state => ({
			|           ^
			60 |             ...state,
			61 |             ...URL2Object(thisSearch),
			62 |           }));`

Another code version also has this error:
```
componentDidUpdate(prevProps) {
        const { search: thisSearch } = this.props.location;
        const { search: prevSearch } = prevProps.location;

        if (thisSearch !== prevSearch) {
          const params = URL2Object(thisSearch);
          this.setState(params);
        }
      }
```

### DIFF
--- a/src/vendor/utils/navigation/index.jsx
+++ b/src/vendor/utils/navigation/index.jsx
@@ -49,6 +49,20 @@ function withNavigation(config) {
         this.updateHistory(change);
       };
 
+      componentDidUpdate(prevProps) {
+        const { search: thisSearch } = this.props.location;
+        const { search: prevSearch } = prevProps.location;
+
+        if (thisSearch !== prevSearch) {
+          // Update state based on current props
+          // eslint-disable-next-line react/no-did-update-set-state
+          this.setState(state => ({
+            ...state,
+            ...URL2Object(thisSearch),
+          }));
+        }
+      }
+
       updateHistory(change) {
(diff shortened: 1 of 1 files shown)


src/vendor/utils/navigation/index.jsx
```diff
@@ -49,6 +49,20 @@ function withNavigation(config) {
         this.updateHistory(change);
       };
 
+      componentDidUpdate(prevProps) {
```
klahnakoski: You should not need to use `componentDidUpdate()` it is used for notifying some *other* parts of your program that properties changed.  

You must investigate how to detect the back button is clicked so this component knows it should re-render.

klahnakoski:
This was driving me crazy, and you fixed it.  thank you very much!

PR #397, opened 2019-04-05
Each series to have many sources

klahnakoski:
The `seriesConfig` now accepts a JSON expression rather than a number of matching properties.  These expressions are used to filter the `PERHERDER.signatures` [code](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/b2b27bc3a79d8ac0c15be437e5992176f9a0d139/src/vendor/perfherder.js#L108) and get the signatures, which are then requested.  

Every series now has an array of `sources`.  Formatting has moved to `src/containers/PerfherderGraphContainer/index.jsx` since it needed to be rewritten anyway.

This removes the need for `perf-goggles`, and uses `/src/vendor/perfherder.js` instead.


klahnakoski:
There may be an argument to be made for using javascript functions, rather than JSON, to encode these rules.  JSON is better because:

1. They are language agnostic, and interpretors are easy to make [here is the javascript one](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/b2b27bc3a79d8ac0c15be437e5992176f9a0d139/src/vendor/jx/expressions.js#L8)
2. JSON can be inspected and manipulated.  For example [I extract the framework out of the expression](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/b2b27bc3a79d8ac0c15be437e5992176f9a0d139/src/vendor/perfherder.js#L77) so that the API stays simple while still keeping the `framework` property very special: Perfherder requires you know the `framework_id` for the suite you are requesting.
3. Composable: Expressions can be compounded. Functions can too, but...
4. JSON expressions can printed, compound javascript methods are difficult to inspect
5. Can be stored in static config files and document stores for sharing

src/vendor/perfherder.js
```diff
@@ -39,7 +40,7 @@ const frameworkCache = {};
 const getFramework = async framework => {
   if (frameworkCache[framework] === undefined) {
     frameworkCache[framework] = (async () => {
-      const url = `${TREEHERDER}/api/project/${repository}/performance/signatures/?${toQueryString(
+      const url = `${TREEHERDER}/api/project/${REPO}/performance/signatures/?${toQueryString(
```
karlht: Is there an important distinction (scope, perhaps?) between `const`s in lower-case and `const`s in UPPER-CASE?

README.md
```diff
@@ -202,7 +202,7 @@ You can also pass an optional `timerange` value, however, this is not necessary
 NOTE: If you want to add Chrome jobs, you have to notice that they're not running on the same platform as the Firefox jobs.
 This means that you need to append `-nightly` to it (e.g. `linux64` -> `linux64-nightly`).
 
-Sample code:
+Sample code (INVALID):
```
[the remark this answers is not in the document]
klahnakoski: Thank you.  I will add it as an issue https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/399

src/vendor/perfherder.js
klahnakoski: It is assumed that uppercase never changes

klahnakoski: ....well, I guess all `const` never change, but all-caps implies application-wide constant.

klahnakoski:
@armenzg May I get your feedback on this PR?

src/quantum/config.js
```diff
@@ -51,75 +52,159 @@ const PLATFORMS = [
     bits: 32,
     os: 'win',
     label: 'Firefox',
-    framework: 10,
-    platform: 'windows7-32',
-    option: 'pgo',
-    project: 'mozilla-central',
+    platform: "win32",
+    seriesConfig: {
+      and: [
```
armenzg: This is incredible. Please go back or find another reviewer.

src/quantum/index.jsx
```diff
@@ -99,12 +163,7 @@ export default class QuantumIndex extends React.Component {
               {
                 label: 'Firefox',
                 seriesConfig: {
-                  extraOptions: ['e10s', 'stylo'],
-                  framework: 1,
-                  platform,
-                  option: 'pgo',
-                  project: 'mozilla-central',
-                  suite: 'tp5o',
+                  and: [performanceFilter, { eq: { suite: 'tp5o' } }],
```
armenzg: This goes from human intelligible to not. No need to defend your position. If you insist on this approach please find another reviewer.

src/utils/chartJs/redashFormatter.js
```diff
@@ -1,10 +1,11 @@
+/* eslint-disable camelcase */
+
 import generateDatasetStyle from './generateDatasetStyle';
 import SETTINGS from '../../settings';
 
 const sortByDate = (a, b) =>
   new Date(b.submission_date) - new Date(a.submission_date);
 const dataToChartJSformat = data =>
-  // eslint-disable-next-line camelcase
```
armenzg: We only disable specific lines rather than the whole file so we don't regress other parts of the file.

klahnakoski:
merged https://github.com/mozilla-frontend-infra/firefox-health-dashboard/commit/c75675815182628851058c95e28ea7a75f2d35cb

klahnakoski:
@helfi92   Good day!  I would appreciate your help to review this PR. I *believe* it is best you start at a high level and point out where I am being unclear so that I may turn them into issues and/or improve the documentation.  

Thank you very much

klahnakoski:
@helfi92 yes, the PR has been merged; we need to show the multitude of platforms for any one series sooner rather than later,  Your review is important: it will help guide the future shape of this code.

src/quantum/config.js
```diff
@@ -203,9 +288,18 @@ const SUITES = {
...
 };
 // ALL PAGE COMBINATIONS
-const TP6_PAGES = selectFrom(SUITES.data)
+const temp = selectFrom(SUITES.data)
   .map(row => Data.zip(SUITES.header, row))
-  .leftJoin('browser', PLATFORMS, 'browser');
-const TP6M_PAGES = TP6_PAGES.where({ browser: 'geckoview' });
+  .toArray();
+
+Log.note("tesmp");
```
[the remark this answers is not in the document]
klahnakoski: Yes typo.  I will remove it

src/quantum/config.js
klahnakoski: I thought you might like this; your original suggestion for this file was to split out the `seriesConfig` from the rest of the properties, and this proves you were right.  But, I also needed a way to declare a number of specific platforms and combinations that this series can be found running on. I can not go back, I need a solution.

Rather than

```
seriesConfig: {
    framework: 10,
    platform: 'windows7-32',
    option: 'pgo',
    project: 'mozilla-central',
}
```

which assumes those properties must match the series, the new format is more explicit: Saying that the properties must be `eq`ual:

```
seriesConfig: { eq: {
    framework: 10,
    platform: 'windows7-32',
    option: 'pgo',
    project: 'mozilla-central',
}}
```

but since there a couple platform combinations, we use some Boolean logic to combine them.

```js
{"seriesConfig":{"and":[
    {"or":[
        {"eq":{"platform":"windows7-32","options":"pgo"}},
        {"eq":{"platform":"windows7-32-shippable","options":"opt"}}
    ]},
    {"eq":{"framework":10,"repo":"mozilla-central"}}
]}}
```

It is unfortunate that the linter spreads this over so many lines.  Here is the Javascript for the same:

```js
  ({ platform, framework, repo, options }) =>
    framework === 10 &&
    repo === 'mozilla-central' &&
    ((platform === 'windows10-64-shippable' && options === 'opt') ||
      (platform === 'windows10-64' && options === 'pgo'))
```

Personally, I find the `{and:[]}` and `{or:[]}` are easier to read than infix Boolean operators with brackets.

I agree, there can be a good argument for avoiding these JSON expressions, and using some specialized format; maybe an array of `seriesConfig` for multiple platforms.  I found that not suitable because it is not general enough: We also have instances of suite names changing (https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/quantum/index.jsx#L257), which multiplied by the number of platforms is [6 signatures](https://treeherder.mozilla.org/perf.html#/graphs?timerange=5184000&series=mozilla-central,1503356,1,1&series=mozilla-central,1521830,1,1&series=mozilla-central,1539226,1,1&series=mozilla-central,1552932,1,1&series=mozilla-central,1902204,1,1&series=mozilla-central,1934089,1,1) (well, it was before the database rewrite).  Plus, there are future expressions; like pulling load time over all the suites so we can calculate aggregates.

src/quantum/index.jsx
klahnakoski: What's happening here is I am defining a common factor `performanceFilter`.

If we were using the original `seriesConfig` structure, I would do something like:

```js
const performanceFilter ={
    extraOptions: ['e10s', 'stylo'],
    framework: 1,
    platform,
    option: 'pgo',
    project: 'mozilla-central'
}
```
 and then using that common factor in each `seriesConfig`:

```js
seriesConfig: {
    ...performanceFilter,
    suite: 'tp5o'
}
```

the same is happening here, only the logic is explicit:

```
seriesConfig: {
    and: [performanceFilter, { eq: { suite: 'tp5o' } }]
}
```

I added the common factor because the linter expands JSON over many lines, and I wanted to avoid the redundancy. The future hope is the whole array of performance tests will be generated from a configuration; multiplying the platforms by the suites, much like the quantum pages are generated now. I did not go that far here so it is easier to review the steps to that new code.

src/quantum/config.js
```diff
@@ -51,75 +52,159 @@ const PLATFORMS = [
...
+        },
+        {
+          eq: {
+            framework: 10,
+            repo: 'mozilla-central',
+          }
+        },
+
+      ]
+    }
```
helfi92: Iterating over `seriesConfig` is a bit complicated with all of the nesting. If the issue is about properly being able to filter through the data, have we thought about using another tool? For example, there's https://github.com/crcn/sift.js which exposes many operators like `$eq`, `$and`, and `$or` which could be used to filter through the array.

klahnakoski:
@helfi92   Thank you for pointing out Sift. I did not know it existed. I see it is based on Mongodb query syntax, while [JSON expressions](https://github.com/mozilla/ActiveData/blob/dev/docs/jx_expressions.md) are based on the [Elasticsearch syntax](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/term-level-queries.html),

I gave [some thought to expression structure while back](https://github.com/mozilla/ActiveData/blob/dev/docs/jx.md#exploring-possible-expression-formats). The difference between Mongo and Elasticsearch is mostly in the ordering of the operators:

MongoDB, with "infix" operators 

    { city: { $exists: false}}
    { state: {$eq: 'MN' }}

vs JSON Expressions, with "prefix" operators

    {exists: 'city'}
    {eq: {state: 'MN'}}

which is the same length, so `seriesConfig` will not get any shorter by using sift.

JSON expressions have two benefits:

1. Easier to write translators and interpreters because all expressions are objects in `{op: params}` form. For example we can [convert JSON expressions to complicated Bugzilla quereies](https://github.com/mozilla/charts/blob/quantum-investigate/quantum/modevlib/Bugzilla.js#L99) easy enough.  I am not sure how the optional `$eq` affects the translator in practice for Mongodb queries: What does `{$eq: {$eq: 'name'}}` mean?
2. Extensible:  `$regex` can do a lot, but there are benefits to `[find](https://github.com/mozilla/ActiveData/blob/dev/docs/jx_expressions.md#find-operator)` and `[prefix](https://github.com/mozilla/ActiveData/blob/dev/docs/jx_expressions.md#prefix-operator)` which usually translate to faster implementations on the receiving side.  Being able to add operators allows us to leverage datastores' unique search API.

PR #413, opened 2019-04-24
fix subbenchmark code, more logging, remove not-used code

klahnakoski:


src/quantum/subbenchmarks.jsx
```diff
@@ -78,7 +79,7 @@ class PerfherderContainer extends Component {
     const { perfherderUrl, data, parentSignature } = await getData({
       suite,
       platform,
-      framework,
+      framework: Number.parseInt(framework, 10),
```
klahnakoski: this is the only important line; the `framework` is arriving from the URL as a string "10", not a number.

PR #416, opened 2019-04-27
x axis to all same size

klahnakoski:
I was hard to compare charts because each would have a different time scale depending on when the min and max x-values occurred.  This is the beginning of ensuring the xaxis points to a formal Domain

### DIFF
--- a/src/vendor/chartJs/README.md
+++ b/src/vendor/chartJs/README.md
@@ -1,0 +1,20 @@
+Proposed Chart API
+==================
+
+Yet another chart API!  The priorities are:
+
+1. No magic labels or prefixes - A surprising number APIs use a concatenation-
+of-names to inject parameter values.  We will perfer the use of objects for compound values.
+2. Use CSS style names - use camelCase names, avoid names that break rule #1
+3. `data` is assumed to be an Array of Objects, or a qb query result.
+4. `value` is can be a property name, or a qb expression
+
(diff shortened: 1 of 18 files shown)


klahnakoski:
@karlht , @helfi92 I added you to review this so you may comment on the "standard" chart property names 

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/416/files#diff-3a7bcdee40d476193a5a7581f5233f51

The long term plan is to use these names to refer to the plethora properties that may be set in charts.  It will serve to separate the dashboard code from the actual charting library implementation. (My previous dashboards used multiple charting libraries to fill the gaps of the others.) It will also allow me to layer features on top of ChartJS, which like all charting libraries, does not do everything.

klahnakoski:
merged https://github.com/mozilla-frontend-infra/firefox-health-dashboard/commit/6a8a5a21a924dac919764ea1a67205ca6e65a33e

src/vendor/chartJs/README.md
```diff
@@ -0,0 +1,20 @@
+Proposed Chart API
+==================
+
+Yet another chart API!  The priorities are:
+
+1. No magic labels or prefixes - A surprising number APIs use a concatenation-
+of-names to inject parameter values.  We will perfer the use of objects for compound values.
```
karlht: Nit: 'perfer' should read 'prefer'

PR #419, opened 2019-04-30
Dead code removal

BrunoDM2943:
Hi. Its my first PR to an open source project.
 Hope to get it right =)


test/components/linearIndeterminate.test.jsx
```diff
@@ -0,0 +1,19 @@
...
+          classes={{
+            root: 'X',
+          }}
+        />
+      )
+      .toJSON();
+
+    expect(tree).toMatchSnapshot();
+  });
+});
```
klahnakoski: Please remove this test, and the `LinearIndeterminate` code it is testing.  It is not used anywhere.

test/components/bugzilla.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+            text: 'Y',
+            margin: 'Z',
+          }}
+        />
+      )
+      .toJSON();
+
+    expect(tree).toMatchSnapshot();
+  });
+});
```
klahnakoski: These two tests cover just one line of code.  Remove the one that is providing no net-new-coverage.

The remaining test will cover just one line.  I looked through the code and the `BugzillaQueryUrl` is used by only by `BugzillaUrlContainer`, which is used only by `src/views/Android/index.jsx`  Can a higher level test provide more coverage?

test/components/linearIndeterminate.test.jsx
BrunoDM2943: Yes sure. I will remove the test. Is it ok to keep the component event though its not used?

test/components/bugzilla.test.jsx
BrunoDM2943: I think thats ok to leave both. Because here we are testing the component it self. Even though another test of a higher component tests it, the correct is to each component have a test. What do you think?

test/components/linearIndeterminate.test.jsx
klahnakoski: Please remove the component too

test/components/bugzilla.test.jsx
klahnakoski: @BrunoDM2943, 

There is this the general issue of scattered implementation: from [this comment](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/288#issue-406416189) I mention this principal:

> Only use indirection if more than one component/method/object must reference it.

This component should be in the same file as `BugzillaUrlContainer`, and the test(s) should be on that higher component instead.   Even better: the `BugzillaUrlContainer` (allong with its unique dependencies) should be in` Android/index.jsx`, because it is the only place it is used; the test should be on the `Android` component.  If you inline this chain of code, you will find that some code paths are impossible, and the lines of code collapses and zero new tests are needed. 

This is one of many instances where this codebase exposes a public entity (`BugzillaQueryUrl`), which really should be private so it does not contribute to the perceived complexity of the code.

Some functions do not need tests: 
* functions that are provably correct - so simple that they are obvious or a single pass through the debugger can show error; in this case a visual inspection is good enough.
* private functions - functions that are part of a larger function

Adding tests to these types of functions makes the the application more difficult to change; they are testing "how" the function is implemented rather than "what" needs to be done. 

I thank you for the effort providing tests, but you stumbled upon a function that really should be cleaned up.

klahnakoski:
Overall please convert this PR from a test-adding PR, to a dead-code removal PR.  You may want to try inlining the BugzillaUrl code.

Thank you!

BrunoDM2943:
@klahnakoski , I removed all the new tests add on this PR and also the `LinearIndeterminate`. I will make another PR later with another tests and maybe rearrange the code from Bugzilla

klahnakoski:
Excellent!  Thank you!

PR #425, opened 2019-05-09
use URL() to build urls

klahnakoski:


src/containers/RedashContainer/index.jsx
```diff
@@ -1,50 +1,24 @@
...
-import fetchJson from '../../utils/fetchJson';
+import { fetchJson } from '../../vendor/requests';
 import { withErrorBoundary } from '../../vendor/errors';
-import { Log } from '../../vendor/logs';
-
-/* eslint-disable camelcase */
+import { selectFrom } from '../../vendor/vectors';
 
-const sortByDate = (a, b) =>
-  new Date(b.submission_date) - new Date(a.submission_date);
```
klahnakoski: replaced with `.sortBy('submission_date')`

src/containers/RedashContainer/index.jsx
```diff
@@ -1,50 +1,24 @@
...
-const dataToChartJSformat = data =>
-  data.map(({ submission_date, value }) => ({
-    x: submission_date,
-    y: value,
-  }));
```
klahnakoski: replaced with `.select({ x: 'submission_date', y: 'value' })`

src/containers/RedashContainer/index.jsx
```diff
@@ -1,50 +1,24 @@
...
-
-    if (!result[key]) {
-      // eslint-disable-next-line no-param-reassign
-      result[key] = [];
-    }
-
-    result[key].push(datum);
-
-    return result;
-  }, {});
```
klahnakoski: the above is replaced with `.groupBy(dataKeyIdentifier)`

src/containers/RedashContainer/index.jsx
```diff
@@ -1,50 +1,24 @@
...
-  const datasets = Object.keys(buckets).map(key => {
-    const datum = buckets[key];
```
klahnakoski: `buckets` was built from an object (`{}`) which is subsequently converted to a sequence {key, datum} pairs above; the intermediate `buckets` is not needed. `.groupBy()` provides this sequence directly.

klahnakoski:
@helfi92   Please provide your opinion on this change.

The `URL()` function packages all the string-concatenation done to construct URLs. The benefits are:

* Less statements to construct a URL - `URL()` does not save on line count, but are expressions rather than multiline statements.  Technically there is nothing stopping us from using expressions with template literals, but I suspect multi line statements were used because...
* Nicer formatting - Long template literals are horribly formatted by the linte; adding line breaks at illogical positions: Logical for string concatenation, but not for URL construction.
* Structured - the path and the parameters are separated for clarity; constant parameters stand out, rather than hide in the string. A structured API to URL construction can make calls simpler than string munging via `stringify` or `toQueryString`.

src/vendor/requests.js
```diff
@@ -0,0 +1,103 @@
...
+
+const URL = ({ path, query, fragment }) => {
+  const paths = toArray(path);
+  const last = path.length - 1;
+  const fullPath =
+    last === 0
+      ? paths[0]
+      : paths
+          .map((p, i) => {
+            let pp = p;
```
helfi92: What does `pp` stand for?

src/vendor/requests.js
```diff
@@ -0,0 +1,103 @@
...
+
+  try {
+    return response.json();
+  } catch (error) {
+    Log.error(`Problem parsing ${response.text()}`);
+    // Log.error("Problem parsing {{text}}", {text: response.text()});
+  }
+};
+
+const URL = ({ path, query, fragment }) => {
```
helfi92: Thoughts on using https://developer.mozilla.org/en-US/docs/Web/API/URL instead?

src/vendor/requests.js
klahnakoski: @helfi92 
Good question!  I will describe my situation:

> Naming things is hard.  There are many cases where I do not wish to name things, and this is one of them; Due to the linter demanding that I can not assign to a function parameter, I must invent a new name for a variable that holds the parameter value. When I must invent a new variable name derived from a short-lived, single-letter variable (in this case `p`), then I simply double it (`pp`).  Sometimes, in nested loop situations, you can see me triple a variable name (`ppp`).

Maybe you have a procedure for name generation under similar situations?

There is another convention I use, and it applies to this situation; I name the return variable `output`.  I could use that name here.

src/vendor/requests.js
klahnakoski: Excellent! Thank you very much for pointing this API out. It does a lot of what I would like to see in the future, like identifying protocol, I appreciate your effort to find a viable alternative.

`URL` has a few drawbacks:

1. `URL` seems to represent slices of the URL string, rather than a serialization-independent API.  For example, `URL.hash` includes the hash (`#`) symbol, which is redundant.  Same with `URL.search`, which includes the question mark (`?`).  `URL.protocol` includes the colon (`:`).
2. `URL` uses unconventional names: [The RFC](https://tools.ietf.org/html/rfc3986) uses "scheme", "query" and "fragment", while `URL` uses "protocol", "search", and "hash".
3. `URL` does not set `query` - One main purpose of my function is to accept a `query` object to build the query part of the URL. The `URL` you point to only allows reading such an object, plus I am not certain that the returned objects are "properly" parsed.
4. `URL` lacks parametric paths - Many RESTful APIs demand segments of the `path` consist of meaningful parameter values.  For example, the `repo` parameter is put in the 3rd step of a 5-step path: `api/project/${repo}/performance/signatures/`.  My class allows parameterization of the `path` by accepting an array of strings, some variable, which will be assembled to a proper path.

src/vendor/requests.js
helfi92: > When I must invent a new variable name derived from a short-lived, single-letter variable (in this case p), then I simply double it (pp).

Ah okay, I understand now. Works for me.

src/vendor/requests.js
```diff
@@ -0,0 +1,103 @@
+/* global fetch */
+import { parse } from 'query-string';
+import { leaves, toPairs } from './vectors';
+import { Log } from './logs';
+import { Data } from './Data';
+import { exists, isArray, isString, toArray } from './utils';
+import strings from './strings';
+
+function fromQueryString(query) {
```
helfi92: At a first peak, I'm uncertain about the format of the query it expects. Does it handle both `?search=..` and `search=`? A comment above this function would be helpful :)

Are there any tests configured for this utility by any chance?

src/vendor/requests.js
```diff
@@ -0,0 +1,103 @@
...
+  const decode = v => {
+    if (isArray(v)) return v.map(decode);
+
+    if (v === null || v === '') return true;
```
helfi92: Not adding braces for if clauses can be somewhat of a footgun if you go back later and add a second statement to the if clause without remembering to add the curly braces, the code will break in unexpected ways. Also, depending on the line length, the linter might put the statement in a newline. I typically always add braces for maintainability.

```js
if (something) {
  doSomething();
}
```

src/vendor/requests.js
```diff
@@ -0,0 +1,103 @@
...
+      url,
+      status: response.status,
+    });
+  }
+
+  try {
+    return response.json();
+  } catch (error) {
+    Log.error(`Problem parsing ${response.text()}`);
+    // Log.error("Problem parsing {{text}}", {text: response.text()});
```
helfi92: I think we can remove the comment.

src/vendor/strings.js
```diff
@@ -83,6 +83,21 @@ const strings = {
     return v;
   },
 
+  trim(value, prefix) {
```
helfi92: Should this utility have tests?

src/vendor/requests.js
klahnakoski: I will add description, and move the tests (which are [here](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/9684d44e481a20e624e0c41e53caaf1162261e99/test/vendor/convert.test.js#L75))

klahnakoski:
@helfi92 thank you!  all good points, and changes made.

Issue #386, opened 2019-04-02
Odd highlight behaviour in chart


klahnakoski:
It seems when highlighting points on charts the Nth point is highlighted on all series

![image](https://user-images.githubusercontent.com/2334429/55409666-6bfc5180-5530-11e9-98c8-a7051d163875.png)

This is not expected

klahnakoski:
fixed with https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/441

PR #431, opened 2019-05-24
Add Battery Stats

klahnakoski:


klahnakoski:
@gmierz please review https://deploy-preview-431--firefox-health-dashboard.netlify.com/power?platform=p2-aarch64&suite=speedometer

klahnakoski:
@gmierz  

* What is `scn-power-idle`?  is that a test, or is it an idle machine?  What does the "scn" stand for?
* Are there other test suites planned?

gmierz:
> 
> 
> @gmierz
> 
>     * What is `scn-power-idle`?  is that a test, or is it an idle machine?  What does the "scn" stand for?

This is an idle test with the browser running on the about:blank page. "scn" stands for scenario which is the test type.

> 
>     * If it is idle, then why is the power consumption higher?

This is because a plain white background (the about:blank page) consumes the most energy on the P2 devices, so it uses more than the speedometer test.

> 
>     * Are there other test suites planned?
>

Yes, we will be adding more tests such as `scn-power-idle-bkg`.

And I think the interface looks good, nice work! I have the following comments:
1) It is unclear what test suite a certain color is a part of (in the data points).
2) The solid dots don't follow each other properly when browsing the data points. When multiple suites are overlayed, the solid dots seems to start at the beginning (time-wise) of each of the time-series.
3) It could be useful to have the ability to overlay test suites across platforms (not necessary).

klahnakoski:
>     1. It is unclear what test suite a certain color is a part of (in the data points).

Fixed.  I will show only one suite at a time now.

>     2. The solid dots don't follow each other properly when browsing the data points. When multiple suites are overlayed, the solid dots seems to start at the beginning (time-wise) of each of the time-series.

Yes, this is a known problem:  https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/386

>     3. It could be useful to have the ability to overlay test suites across platforms (not necessary).

@gmierz 

I agree.  It would be nice if the user can choose how to show the four dimensions (`Platform` x `Browser` x `Suite` x `Test`) on the 2-dimentional page (`series-in-chart`, `chart-on-page`).  Right now the `series-in-chart` is showing `Browser`, which is uninteresting.  Maybe it should show `Platform`, and leave the user to select `Browser` at the top?

klahnakoski:
@gmierz I pushed a new version that shows all platforms.  Which do you like?

 https://deploy-preview-431--firefox-health-dashboard.netlify.com/power?browser=geckoview&suite=speedometer

src/components/DashboardPage/index.jsx
```diff
@@ -30,15 +38,25 @@ const DashboardPage = ({ classes, children, title, subtitle }) => (
...
+        <Link to="/quantum/64" title="Quantum 64 Performance">
+          <DesktopIcon />
+        </Link>
+        <Link to="/power" title="Power Usage">
+          <BatteryIcon />
+        </Link>
+        <a
+          href="https://github.com/mozilla-frontend-infra/firefox-health-dashboard/"
+          title="Source Code">
+          <CodeIcon />
```
klahnakoski: above code is showing icons in top left of screen:

image

src/components/Section/index.jsx
```diff
@@ -29,14 +30,15 @@ const Section = ({ classes, children, subtitle, title, more }) => (
...
+          {subtitle && <span className={classes.subtitle}>{subtitle}</span>}
           {more && (
             <span>
-              {' ('}
-              <a href={more}>more</a>
-              {')'}
+              {' '}
+              <a href={more} title="show details">
+                <DetailsIcon />
+              </a>
```
klahnakoski: Use icon instead of word

src/components/TP6mAggregate.jsx
```diff
@@ -287,12 +287,11 @@ class TP6mAggregate_ extends Component {
                                 platform,
                               },
                             })}
+                            title="show details"
                             target="_blank"
-                            rel="noopener noreferrer"
-                            style={{ fontSize: '0.9em' }}>
-                            {'details'}
+                            rel="noopener noreferrer">
+                            <DetailsIcon />
```
klahnakoski: Use icon instead of word

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -20,15 +20,15 @@ const ALLOWED_TREEHERDER_TIMERANGES = [1, 2, 7, 14, 30, 60, 90].map(
...
 
   return {
-    reverse: !lowerIsBetter,
     tooltips: {
       enabled: false,
     },
     // Start using: chartSchema.md
-    'axis.y.label': lowerIsBetter ? 'Duration' : 'Score',
+    'axis.y.label': unit,
+    'axis.y.reverse': !lowerIsBetter,
```
klahnakoski: fixing the perherder data to emit `unit`, and we can use it here

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -132,10 +132,6 @@ const styles = () => ({
     padding: '.2rem .3rem .3rem .3rem',
     margin: '0 1rem 0 0',
   },
-  linkIcon: {
-    marginLeft: '0.2rem',
-    marginBottom: -5,
-  },
```
klahnakoski: remove CSS where it is obvious we do not need it

src/containers/RedashContainer/index.jsx
```diff
@@ -81,32 +82,29 @@ class RedashContainer extends Component {
...
+              {title}{' '}
+              <a
+                href={redashQueryUrl}
+                target="_blank"
+                rel="noopener noreferrer"
+                title="go to Redash query">
+                <LinkIcon />
+              </a>
+            </div>
+          }
```
klahnakoski: Use icon instead of words

src/index.css
```diff
@@ -12,19 +12,18 @@
...
-  height:100%
+ html, body {
+  overflow-x: hidden;
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  margin: 0;
+  padding: 0;
```
klahnakoski: There is zero-pixel overflow, which triggered the horizontal scrollbar, which is now gone.

src/index.css
```diff
@@ -46,49 +45,10 @@
...
-  text-align:right
- }
- .dashboard-title .dashboard-source a:last-child {
-  display:block
- }
- .dashboard-title .dashboard-link {
-  padding:.53572rem .375rem;
-  color:#00a7e0;
-  display:none
- }
```
klahnakoski: more CSS removal

src/power/config.js
```diff
@@ -0,0 +1,74 @@
...
+    ['geckoview', 'Geckoview', 'speedometer', 'Speedometer', {eq: {framework: 10, repo: 'mozilla-central', suite: ['raptor-speedometer-geckoview-power']}}],
+    ['refbrow', 'Reference Browser', 'speedometer', 'Speedometer', {eq: {framework: 10, repo: 'mozilla-central', suite: ['raptor-speedometer-refbrow-power']}}],
+    ['fenix', 'Firefox Preview', 'speedometer', 'Speedometer', {eq: {framework: 10, repo: 'mozilla-central', suite: ['raptor-speedometer-fenix-power']}}],
+    ['fennec', 'Fennec', 'speedometer', 'Speedometer', {eq: {framework: 10, repo: 'mozilla-central', suite: ['raptor-speedometer-fennec-power']}}],
+
+  ],
+};
+const COMBOS = COMBO_TABLE.data.map(row => Data.zip(COMBO_TABLE.header, row));
+
+export { TESTS, COMBOS, PLATFORMS };
```
klahnakoski: This is the config file to drive the `/power` page

src/power/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+    id: 'suite',
+    label: 'Suite',
+    defaultValue: 'speedometer',
+    options: selectFrom(COMBOS)
+      .groupBy('suiteLabel')
+      .map(([v]) => ({ id: v.suite, label: v.suiteLabel })),
+  },
+];
+
+export default withNavigation(nav)(withStyles(styles)(Power));
```
klahnakoski: The actual `/power` page

src/quantum/Quantum.jsx
```diff
@@ -581,11 +582,10 @@ export default class QuantumIndex extends React.Component {
...
                 {more && (
                   <span>
-                    {' ('}
-                    <a href={more} style={{ fontSize: '0.8em' }}>
-                      more
+                    {' '}
+                    <a href={more} title="show details">
+                      <DetailsIcon />
                     </a>
-                    {')'}
```
klahnakoski: Use icon instead of word

src/quantum/config.js
```diff
@@ -47,7 +44,6 @@ const CONFIG = {
       title: 'MotionMark Animometer',
       secondLink:
         '/quantum/10/windows7-32/raptor-motionmark_animometer-firefox/pgo',
-      secondTitle: 'Breakdown',
```
klahnakoski: we no longer need the word because we use icon

src/quantum/widget.jsx
```diff
@@ -35,14 +36,14 @@ export default class Widget extends React.Component {
...
-        $secondTitle = this.props.secondTitle;
-      }
+    if (this.props.secondLink) {
+      $secondTitle = (
+        <Link to={this.props.secondLink} title="show details">
+          <DetailsIcon />
+        </Link>
+      );
+    } else {
+      $secondTitle = '';
```
klahnakoski: Use icon instead of word

src/utils/icons.jsx
```diff
@@ -0,0 +1,60 @@
...
+  TimerIcon,
+  BatteryIcon,
+  VideoIcon,
+  DesktopIcon,
+  AndroidIcon,
+  DetailsIcon,
+  GithubIcon,
+  CodeIcon,
+  HomeIcon,
+};
```
klahnakoski: All icons are here, plus some styling because they are badly behaved

src/vendor/chartJs/utils.js
```diff
@@ -74,7 +75,7 @@ const generateOptions = (rawOptions = {}, data) => {
     {
       ticks: {
         beginAtZero: true,
-        reverse,
+        reverse: Data.get(options, 'axis.y.reverse'),
```
klahnakoski: FIX: be sure to pass in the `reverse`

src/vendor/logs.js
```diff
@@ -25,25 +25,28 @@ function parseStack(stackString) {
...
+            fileName: parts[2],
+            line: parts[3],
+            column: parts[4],
+          };
+        })
+        .exists()
+        .first()
+    )
+    .exists()
+    .toArray();
```
klahnakoski: A little more scrubbing of the stack traces

src/vendor/perfherder.js
```diff
@@ -40,10 +40,23 @@ const getFramework = async combo => {
...
-          let lowerIsBetter = true;
+          const cleanTest = (() => {
+            if (missing(test)) return null;
+
+            if (test === suite) return null;
+
+            if (test.startsWith(suite)) return test.slice(suite.length + 1);
+
+            return test;
+          })();
```
klahnakoski: Test names are can come in many different flavors: We scrub them.

src/vendor/perfherder.js
```diff
@@ -40,10 +40,23 @@ const getFramework = async combo => {
...
+          let lowerIsBetter = lower_is_better;
+          let unit = 'Score';
 
           if (lower_is_better === undefined) {
-            if (
+            if (suite.endsWith('-power')) {
+              lowerIsBetter = true;
+              unit = 'mAh';
+            } else if (
```
klahnakoski: Markup `power` tests with correct unit

src/vendor/utils/links.jsx
```diff
@@ -0,0 +1,21 @@
...
+    return (
+      <a href={to} title={title}>
+        {children}
+      </a>
+    );
+  }
+}
+
+/* eslint-disable-next-line import/prefer-default-export */
+export { Link };
```
klahnakoski: The builtin `Link` can not be used outside a router, which then breaks the tests.  This `Link` does not have that problem.

src/views/Android/TP6m.jsx
```diff
@@ -130,69 +127,65 @@ class TP6M extends React.Component {
       .first().label;
 
     return (
-      <div className={classes.body}>
```
klahnakoski: remove more styling

src/views/Home/index.jsx
```diff
@@ -35,10 +36,18 @@ const Home = ({ classes }) => (
...
+      </Link>
+      <Link to="/quantum/32">
+        <DesktopIcon /> Quantum 32bit
+      </Link>
+      <Link to="/quantum/64">
+        <DesktopIcon /> Quantum 64bit
+      </Link>
+      <Link to="/battery">
+        <BatteryIcon /> Power Usage
+      </Link>
```
klahnakoski: Add link to `/battery` and add icons

src/routes.jsx
```diff
@@ -18,13 +18,14 @@ const Routes = () => (
...
       <Route path="/android/graph" component={NimbledroidGraphPage} />
+
       <Route path="/android" component={AndroidPage} />
+      <Route path="/power" component={Power} />
       <Route
         path="/quantum/:framework/:platform/:suite/:option"
         component={Subbenchmark}
       />
       <Route path="/quantum/:bits" component={Quantum} />
-      <Route path="/js-team" component={JsTeam} />
```
klahnakoski: removed `js-team` page because it was never done

klahnakoski:
@sarah-clements   Please review this PR. 

There important part is the addition of the `power` detail page:

* [power/config.js](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/431/files#diff-a4b283cd94bdf6fe52515d6d43a065b4)
* [power/index.jsx](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/431/files#diff-9259b06913a0987519c654f50685f409)

Some other notes:

1. The rest is CSS cleanup, and adding clickable icons for menus, tabs, and drill-down.  I added comments to explain the various changes.  
2. I must still discuss with management about what they want to see on the main dashboard: That will be done in another PR.  

Thank you

klahnakoski:
@karlht  Your review is also appreciated

gmierz:
@klahnakoski I prefer the second one that let's us see all the platforms for a given test suite. But perf-sheriffs might prefer the original interface so it might be a good idea to ask igoldan about it as well. I think this depends on how we will look at the data to locate regressions, and this will constantly change from one issue to another.

klahnakoski:
> @klahnakoski I prefer the second one that let's us see all the platforms for a given test suite. But perf-sheriffs might prefer the original interface so it might be a good idea to ask igoldan about it as well. I think this depends on how we will look at the data to locate regressions, and this will constantly change from one issue to another.

@gmierz   This dashboard is not meant for performance sheriffing, nor for developers; The audience is managers: They will be providing their perspective this week, so it may change again.

src/components/DashboardPage/index.jsx
sarah-clements: Icons look nice but what does the <> mean? Perhaps you could find a clearer icon?

src/power/config.js
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable */
+
+import { Data } from '../vendor/Data';
+
+const TESTS = [
+  {
+    id: 'cpu',
+    label: 'CPU',
+    filter: { eq: { test: 'cpu' } },
```
sarah-clements: It seems like for something like an id (presumably used for internal purposes) you can just use the label?

src/quantum/widget.jsx
sarah-clements: I'm curious why you're using a $ in front of secondTitle.

src/utils/icons.jsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import LinkIco from '@material-ui/icons/Launch';
+import HomeIco from '@material-ui/icons/Home';
+import TimerIco from '@material-ui/icons/Timer';
+import BatteryIco from '@material-ui/icons/BatteryChargingFull';
+import VideoIco from '@material-ui/icons/Videocam';
+import DesktopIco from '@material-ui/icons/DesktopWindows';
+import AndroidIco from '@material-ui/icons/PhoneAndroid';
+import DetailsIco from '@material-ui/icons/MoreHoriz';
+import CodeIco from '@material-ui/icons/Code';
```
sarah-clements: It's odd to leave the 'n' off of Icon here. I think it's better to have a longer name that's clearer, rather than a short vague one :) So `materialLinkIcon` or or rename your icon components to something like `LinkIconStyle`.

However, since some of the styling is the same, couldn't you create a function that returns the div with the style and pass in the Icon element or prop (for `LinkIcon` and `DetailsIcon`)?

src/quantum/widget.jsx
klahnakoski: That is curious; I did not even notice!  I was simply using the variable that was used before.  I will fix this.

src/power/config.js
klahnakoski: In general, I separate the machine-friendly `id` from the humane `label` because they too-often diverge. 

Humans like spaces and capitalization, and sometimes like names to change, while machines like consistency (eg all lowercase, or not changing) and standardization (eg use third party names, like from perfherder).  For example; I have no idea what `proportional` means: I will learn shortly, and it will be easy to change.

src/utils/icons.jsx
klahnakoski: Yes, since these names do not escape this module, I did not care to invent a better name.  The linter gets angry with any names with prefix or suffix of underscore (`_`), so instead of adding a character, I removed one. I will add `Internal` as a suffix instead.

> However, since some of the styling is the same, couldn't you create a function that returns the div with the style and pass in the Icon element or prop (for `LinkIcon` and `DetailsIcon`)?

I think what you suggest will be just as much code as what I have done.

sarah-clements: I was thinking something like this :) But maybe not worth for two icons.
```
const IconWithStyle = (icon) => (
  <div
    style={{
      display: 'inline-block',
      margin: '-1rem 0',
      verticalAlign: 'middle',
    }}>
    {icon}
  </div>
);
```

klahnakoski: @sarah-clements 

I have a question about your suggestion:  Is styling an outer `div` that contains an inner component the same as styling the inner component? Since CSS is capable of insane inter dependencies; I am worried that more html tags, means more CSS styling opportunities, which results in more CSS rat-nest logic. This assumes I can even get the inner component to behave by styling the outer `div`.

On the other hand I have not seen this nightmare scenario in real life, so I wonder if there is some React magic happening here.

sarah-clements: I think this suggestion would be the same as embedding the component - you're styling the outer container and then any props passed directly to the component before its passed in would apply to that component's container (or whatever its used for).

PR #433, opened 2019-05-29
Add time domain 

klahnakoski:


klahnakoski:
@helfi92 

It is best to view this PR after the first commit, which moved a bunch of files around:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/433/files/e7f14b4de375feefe5f39ca2ca967a16af9a34b1..514717812d26822598c1879e60cbe2dce58c2b73

src/android/TP6m.jsx
```diff
@@ -36,7 +37,9 @@ class TP6M extends React.Component {
   }
 
   async componentDidMount() {
-    const { test, platform } = this.props;
+    const { test, platform, past, ending } = this.props;
+    // BE SURE THE timeDomain IS SET BEFORE DOING ANY await
+    const timeDomain = new TimeDomain({ past, ending });
```
klahnakoski: This is a common theme: Converting the `past`, `ending` pair into a TimeDomain object

src/android/TP6m.jsx
```diff
@@ -49,7 +52,7 @@ class TP6M extends React.Component {
       sites,
       tests,
       platforms: selectFrom(PLATFORMS).where({ platform }),
-      timeDomain: DEFAULT_TIME_DOMAIN,
+      timeDomain,
```
klahnakoski: The good part is we will be removing the use of `DEFAULT_TIME_DOMAIN`

src/android/TP6m.jsx
```diff
@@ -106,16 +109,19 @@ class TP6M extends React.Component {
 
   async componentDidUpdate(prevProps) {
     if (
-      this.props.test === prevProps.test &&
-      this.props.platform === prevProps.platform
-    )
+      ['test', 'platform', 'past', 'ending'].every(
+        v => this.props[v] === prevProps[v]
+      )
+    ) {
```
klahnakoski: With more properties (`past` and `ending`) that can be changed, we much check them to determine if we reload

src/android/index.jsx
```diff
@@ -80,7 +82,7 @@ class Android extends Component {
         <Section
           title="Raptor (TP6m)"
           more="/android/tp6m?test=cold-loadtime&platform=geckoview-p2-aarch64">
-          <TP6mAggregate />
+          <TP6mAggregate timeDomain={timeDomain} />
```
klahnakoski: We must pass the `timeDomain` into many more objects and templates.

src/quantum/TP6.jsx
```diff
@@ -80,14 +79,14 @@ const nav = [
...
   },
-
   {
     type: Picker,
     id: 'bits',
     label: 'Bits',
     defaultValue: 64,
     options: [{ id: 32, label: '32 bits' }, { id: 64, label: '64 bits' }],
   },
+  ...timePickers,
```
klahnakoski: These `timePickers` are defined in separate file...

src/utils/timePickers.js
```diff
@@ -0,0 +1,34 @@
...
+  {
+    type: TimePicker,
+    id: 'ending',
+    label: 'Ending',
+    defaultValue: todayText,
+    options: [{ id: todayText, label: 'Today' }],
+  },
+];
+
+export { timePickers, todayText };
```
klahnakoski: This is config for the time pickers; which is the same in all the components that use them.

src/vendor/Template.js
```diff
@@ -1,11 +1,17 @@
...
 import strings from './strings';
 
 let expandAny = null;
 
+class Template {
+  constructor(template) {
+    this.template = template;
+  }
+}
+
```
klahnakoski: I have been writing code that has import cycles: This mostly, works, but it will eventually prevent tests from passing as node gets too confused.  It also spews warnings in the browser console.  I had to fix them, which means moving definitions around.

src/vendor/Template.js
```diff
@@ -15,15 +21,15 @@ function expandLoop(loop, namespaces) {
   const { from, template, separator } = loop;
 
   if (!isString(from)) {
-    Log.error('expecting from clause to be string');
+    Template.log.error('expecting from clause to be string');
```
klahnakoski: The `logs.js` and `Template.js` imported each other.  Now, I will have `logs.js` import `Template.js` and then set `Template.log` to the `Log` object

src/vendor/components/navigation/DurationPicker.jsx
```diff
@@ -0,0 +1,146 @@
...
+      })
+    ),
+    PropTypes.instanceOf(ArrayWrapper),
+  ]).isRequired,
+  width: PropTypes.string,
+};
+
+const DurationPicker = withStyles(styles)(DurationPickerPre);
+
+export { DurationPicker, QUERY_TIME_FORMAT };
```
klahnakoski: This DurationPicker uses the slider to pick one of the available durations

src/vendor/components/navigation/TimePicker.jsx
```diff
@@ -0,0 +1,92 @@
...
+    selectFrom(options)
+      .select('id')
+      .includes(value)
+  )
+    return value;
+
+  options.push({
+    id: value,
+    label: Date.newInstance(value).format('MMM dd, yyyy'),
+  });
```
klahnakoski: Above:  The TimePicker is different than the other pickers. Other pickers will confirm the URL has a legit value, and if not then use the default value. For the TimePicker; if you put an ending date in the URL, it will be used as an option.

src/vendor/components/links.jsx
```diff
@@ -17,5 +17,4 @@ class Link extends React.Component {
   }
 }
 
-/* eslint-disable-next-line import/prefer-default-export */
-export { Link };
+export { Link }; // eslint-disable-line import/prefer-default-export
```
helfi92: Would `export default { Link };` work?

src/vendor/components/navigation/DurationPicker.jsx
```diff
@@ -0,0 +1,146 @@
...
+        width,
+      };
+
+      return;
+    }
+
+    const index = options.findIndex(({ id }) => id === defaultValue);
+
+    this.state = { index, range, options, event: { target: {} }, width };
+    this.handleDragEnd();
```
helfi92: This method will end up callling `this.setState` which we don't want in the constructor. `this.setState` causes a re-render of the component and all of its children. We don't want that in the constructor because the component hasn't rendered yet.

src/vendor/components/navigation/Picker.jsx
```diff
@@ -0,0 +1,86 @@
...
+  },
+});
+
+class Picker extends React.Component {
+  constructor(props) {
+    super(props);
+
+    const { value } = this.props;
+
+    this.state = { value };
```
helfi92: You may want to use `getDerivedStateFromProps` instead here. The danger here is that if the props on the component are changed without the component being recreated, the new prop will never be displayed because the constructor will never update the current state of the component. The initialization of state from props only runs when the component is first created.

src/vendor/components/navigation/TimePicker.jsx
```diff
@@ -0,0 +1,93 @@
...
+  },
+});
+
+class TimePicker extends Component {
+  constructor(props) {
+    super(props);
+
+    const { value } = this.props;
+
+    this.state = { value };
```
helfi92: See other comment in the Picker component.

src/vendor/components/navigation/index.jsx
```diff
@@ -149,9 +137,9 @@ function withNavigation(config) {
 
     const styles = () => ({
       root: {
-        display: 'flex',
+        // display: 'flex',
```
helfi92: Remove this line.

src/vendor/utils.js
```diff
@@ -212,6 +212,35 @@ function concatField(...many) {
...
+  }
+
+  return val.constructor === OBJECT_CONSTRUCTOR || val instanceof Data;
+}
+
+function sleep(ms) {
+  return new Promise(resolve => {
+    setTimeout(resolve, ms);
+  });
+}
```
helfi92: A couple of questions re: `isData` and `sleep`
* what is `isData` checking for?
* why do we need to check the constructor property?
* why `sleep`?

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -136,21 +135,31 @@ const styles = () => ({
...
+  }
 
   async componentDidMount() {
-    const { series, style, reference } = this.props;
+    const { series, style, reference, timeDomain } = this.props;
 
-    try {
-      this.setState({ isLoading: true });
+    this.setState({ isLoading: true });
+    await sleep(0); // ALLOW UI TO UPDATE
```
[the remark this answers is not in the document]
klahnakoski: Too much is happening synchronously!  When all the data is loaded (no network calls are made), a call to update the page takes a couple of seconds for a long calculation to run.  This locks up the UI for that time; preventing redrawing.  `await sleep(0)` adds a "yield" to the execution so that UI elements can update; so that `render()` is executed; before this long calculation.

src/vendor/utils.js
klahnakoski: @helfi92 

`isData` is checking for `Data` objects: 

> Represent JSON object: map from string keys to values

Plain objects, objects created with `{}`, can act like `Data`.  So they are included.

`sleep()` is used to insert delays into an asynch function. `sleep(0)` is intended to let other promise chains execute.

src/android/TP6m.jsx
```diff
@@ -106,16 +109,19 @@ class TP6M extends React.Component {
...
       return;
+    }
 
     this.componentDidMount();
```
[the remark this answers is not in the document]
klahnakoski: `componentDidUpdate` and `componentDidMount` are long running `async` functions: They setup a promise chain, and return immediately.  They can be safely called at any time to update the component state.

I could create a third function, `async getDataAndUpdateComponent()`, and change `componentDidMount` to be  

```javascript
async componentDidMount() {
    return getDataAndUpdateComponent();
}
```

and change `componentDidUpdate` to use `getDataAndUpdateComponent()` too.  But, adding `getDataAndUpdateComponent` requires an extra named function for no benefit.

src/containers/BugzillaUrlContainer/index.jsx
```diff
@@ -11,6 +11,23 @@ const styles = {
...
   },
 };
+const BugzillaUrl = ({ bugCount, text, url }) => (
+  <a href={url} target="_blank" rel="noopener noreferrer">
+    {!bugCount && text}
+    {bugCount && (
+      <Badge color="primary" badgeContent={bugCount}>
+        {text}
+      </Badge>
+    )}
```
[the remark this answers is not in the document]
klahnakoski: This was part of the file moving: Specifically, moving code for this component into its parent file for less files.

src/containers/BugzillaUrlContainer/index.jsx
```diff
@@ -11,6 +11,23 @@ const styles = {
...
+  </a>
+);
+
+BugzillaUrl.propTypes = {
+  bugCount: PropTypes.number,
```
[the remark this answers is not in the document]
klahnakoski: This is not my code; it is a side effect of the file moves. I think that will change behavior; I can not be sure if the above code is looking for `null` or for a zero (`0`), or if it matters.  I can not know until I run it through a debugger.

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -136,21 +135,31 @@ const styles = () => ({
...
+      Log.error('expecting a time range');
+    }
+
+    this.state = {
+      data: null,
+      tooltipModel: null,
+      tooltipIsLocked: false,
+      canvas: null,
+      isLoading: false,
+    };
```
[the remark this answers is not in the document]
klahnakoski: I think putting initialization inside and outside the `constructor` is confusing.  I will keep this in mind; I could remove the value check (by using PropTypes), which allows me to remove the constructor, and then use the technique you suggest.

src/vendor/components/navigation/Picker.jsx
klahnakoski: @helfi92 

I must think about that.  The intent is to set a default value for the Picker.  I can not foresee the `props` ever changing, and even if they did I can not imagine how that would happen, or what it would mean.  The Picker is meant to be the authority on the `value`; having another authority (presumably something that changes `props`) would be interesting.

Maybe the authority of `value` is the navigation component that holds all the Pickers; then the data flows from parent to children; then I must be concerned with `props` changing.

I have a question:  There appears to be two ways to deal with changing `props`: One is with `componentWillUpdate` and the other is ensure the parent sets the `key` attribute [example](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/e929da1c12e91e5ca97dddbc8ba3dea8b620b6ea/src/android/TP6m.jsx#L167).  The former only seems useful when there is no parent, or I do not want the parent to manage child state.  The latter is generally preferred because it lets me assume `props` will not change.  

You have some wise words about when either technique is preferred?

Thank you

helfi92: `componentWillUpdate` was actually [deprecated](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount) in recent versions of React so I wouldn't recommend using them. The React docs has some [excellent documentation](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html) about when what to use. They also tackle the issue about when to use the `key` prop. Have a look and let me know if you still had questions. Happy to help.

src/android/TP6m.jsx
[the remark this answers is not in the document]
helfi92: I think the benefit is that it will be more readable for React users. It should be okay to do the alternative you are suggesting.

From https://reactjs.org/docs/react-component.html#componentdidmount

> componentDidMount() is invoked immediately after a component is mounted (inserted into the tree).

Calling this lifecycle method again will confuse users reading the docs because React users will think that this method is only called once per initialization whereas here we are calling it multiple times.

src/vendor/Template.js
```diff
@@ -156,16 +169,10 @@ function expand(template, parameters) {
...
-  }
-}
-
-Template.expand = expand;
+Template.prototype.expand = function expand(values) {
+  return expando(this.template, values);
+};
 
-export default Template;
+export { Template }; // eslint-disable-line import/prefer-default-export
```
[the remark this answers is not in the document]
klahnakoski: That export demands the import be changed to match; the named imports will no longer work.

I do not know why ES6 has two different types of imports, it it an unnecessary complication.  I do not like guessing if I should be importing a `default` export or a named export.

Since the `default` exports are less flexible, I try to stick with named exports for uniformity.

src/containers/PerfherderGraphContainer/index.jsx
```diff
@@ -255,17 +264,23 @@ class PerfherderGraphContainer extends Component {
...
-  series: PropTypes.arrayOf(
-    PropTypes.shape({
-      label: PropTypes.string.isRequired,
-      filter: PropTypes.shape({}).isRequired,
-      options: PropTypes.shape({
-        includeSubtests: PropTypes.bool,
-      }),
-    })
-  ),
+  series: PropTypes.oneOfType([
```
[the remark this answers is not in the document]
klahnakoski: @helfi92  What duplication did you see?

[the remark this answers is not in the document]
helfi92: I don't see a duplication right now but for future views that may take a serie as a prop, it would be nice to import that prop from a `prop-types` files instead of having to re-define it. I assumed a serie is a general data structure that may be consumed by future components.

src/containers/PerfherderGraphContainer/index.jsx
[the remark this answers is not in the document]
helfi92: Instead of setting state in `componentDidMount`, you could do it in the constructor. This will make sure there is no extra render. This will also make sure the UI updates with the loading screen.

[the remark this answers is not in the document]
helfi92: https://reactjs.org/docs/react-component.html#componentdidmount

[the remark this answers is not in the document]
klahnakoski: I think this code belongs here: It has two main modes; the first time it is run (for the first instance of this component type) it takes a long time to run because lots of data is fetched from Treeherder.  So, I would like this component to show during that time; showing spinners in place of charts.  The subsequent runs (I guess for other instances of this same component), there is a good chance network requests are NOT made; and all promises resolve immediately. This method still takes long to run; not as long as the first run; but long enough to lock up the UI as the long promise chain selfishly resolves; and prevents other promise chains from executing; causing the lockup.  In this second case, I still want the ui to render spinners in place of the charts while we wait, but that will not happen unless I break up my promise chain to let others execute and actually let the UI update.  This single `sleep(0)` helps a lot, but more should be added so the ui is allowed to be responsive.

PR #467, opened 2019-07-04
Add Power Summary to Android Dash

klahnakoski:


klahnakoski:
https://deploy-preview-467--firefox-health-dashboard.netlify.com/android

davehunt:
Could we retitle this from "Speedometer CPU usage" to "Speedometer CPU power usage" to indicate that this is related to the power measurements rather than CPU utilisation?

Issue #472, opened 2019-07-10
Update platform name for Firefox Preview tp6m results


davehunt:
The platform named used when submitting the Firefox Preview results was recently fixed (https://github.com/mozilla-mobile/fenix/commit/00271cc20de59682603f550fe23cac8dba9ce18a) to be consistent with other targets. It is now android-hw-p2-8-0-android-aarch64 instead of android-hw-p2-8-0-aarch64. We need to update the graphs to include this additional signature. It looks like we need to update the filter here:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/5c9f3350116bfdba0c6c7728c51d64b6c057eeda/src/quantum/config.js#L245

PR #476, opened 2019-07-10
Change platform filter for Firefox Preview p2

arnold-iakab:
[issue 472](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/472)

### DIFF
--- a/src/quantum/config.js
+++ b/src/quantum/config.js
@@ -240,16 +240,20 @@ const PLATFORMS = [
     browser: 'fenix',
     label: 'Firefox Preview p2 (aarch64)',
     platform: 'fenix-p2-aarch64',
-    platformFilter: {
-      and: [
-        { eq: { platform: 'android-hw-p2-8-0-aarch64' } },
-        { eq: { platform: 'android-hw-p2-8-0-android-aarch64' } },
+    platformFilter: {and: [
+        {or: [
+          {eq: {
+            platform: 'android-hw-p2-8-0-aarch64'
+          }},
+          {eq: {
+            platform: 'android-hw-p2-8-0-android-aarch64'
+          }},
+        ]},
         {eq: {
           framework: 10,
           repo: 'fenix',
         }},
-      ],
-    },
+      ]},
   },
 ];
 const TP6_TESTS_DATA = [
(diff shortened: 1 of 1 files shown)


src/quantum/config.js
```diff
@@ -242,7 +242,7 @@ const PLATFORMS = [
     platform: 'fenix-p2-aarch64',
     platformFilter: {
       and: [
-        { eq: { platform: 'android-hw-p2-8-0-aarch64' } },
+        { eq: { platform: 'android-hw-p2-8-0-android-aarch64' } },
```
davehunt: I suspect this will only show the new data. Can we include both?

armenzg: @davehunt is this better?

davehunt: I don't know the syntax, but should this be an `or` instead of an `and`?

armenzg: Probably. I don't know the syntax either.

I will leave it for @arnold-iakab to give it a try.

arnold-iakab:
@armenzg I checked if other platforms aggregate data and found an example and applied the same syntax to our case and it seems to work.

src/quantum/config.js
```diff
@@ -240,16 +240,20 @@ const PLATFORMS = [
...
-        { eq: { platform: 'android-hw-p2-8-0-android-aarch64' } },
+    platformFilter: {and: [
+        {or: [
+          {eq: {
+            platform: 'android-hw-p2-8-0-aarch64'
+          }},
+          {eq: {
+            platform: 'android-hw-p2-8-0-android-aarch64'
+          }},
+        ]},
```
klahnakoski: Instead of 

        {or: [
          {eq: {
            platform: 'android-hw-p2-8-0-aarch64'
          }},
          {eq: {
            platform: 'android-hw-p2-8-0-android-aarch64'
          }},
        ]}

you can use 

          {in: {platform: [
              'android-hw-p2-8-0-aarch64', 
              'android-hw-p2-8-0-android-aarch64'
          ]}}

https://github.com/mozilla/ActiveData/blob/dev/docs/jx_expressions.md#in-operator

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/vendor/jx/expressions.js#L167...L173

I will add more comments

Issue #483, opened 2019-07-12
Clarifying resolutions in media playback tests 


arnold-iakab:
Add "p" to the values in the media playback graphs so that the resolution is not interpreted as a value.

Issue #482, opened 2019-07-12
Add media playback tooltips to explain results (green, yellow, red)


arnold-iakab:
The app shows the colors based on the following logic:
- Green when all speeds have loss <= 1
- Yellow when all speeds except _1x_ have loss > 1
- Red when loss for _1x_ speed is > 1

Based on this, we want the _tooltip_ to display: 
- for Green - "One or less dropped frames at all playback speeds"
- for Yellow - "x dropped frames at x playback speed" (show the values for the slowest playback speed that exceeds 1 dropped frame)
- for Red - "x dropped frames at 1x playback speed"

arnold-iakab:
Green when loss for all speeds is  <=  1

PR #488, opened 2019-07-15
Media playback tooltips

arnold-iakab:
@armenzg This PR is ready. It contains two commits that fix [Issue 482](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/482) and [Issue 483](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/483)

I've also added a new npm package, [react-simple-tooltip](https://www.npmjs.com/package/react-simple-tooltip), to the project that I've used for this task.

### DIFF
--- a/src/playback/summary.jsx
+++ b/src/playback/summary.jsx
@@ -57,11 +57,11 @@ const styles = {
   },
 };
 const SPECIAL_SIZES = [
-  { id: '480p30', label: '480' }, // .4M pixels
-  { id: '720p60', label: '720' }, // 1M pixels
-  { id: '1080p60', label: '1080' }, // 2M pixels
-  { id: '1440p60', label: '1440' }, // 4M pixels
-  { id: '2160p60', label: '2160' }, // 8M pixels
+  { id: '480p30', label: '480p' }, // .4M pixels
+  { id: '720p60', label: '720p' }, // 1M pixels
+  { id: '1080p60', label: '1080p' }, // 2M pixels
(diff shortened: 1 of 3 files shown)


src/playback/summary.jsx
```diff
@@ -153,6 +155,30 @@ class PlaybackSummary extends React.Component {
     await this.update();
   }
 
+  tooltipContent(score, encoding, platform, size) {
+    if (score === lookupType['0'])
```
klahnakoski: instead of 

    if (score === lookupType['0'])

use

    if (score === 'pass')

You may want to declare constants,

```
const PASS = 'pass';
const BAD = 'bad';
const FAIL = 'fail';
const lookupType = {
  '0': PASS,
  '1': BAD,
  '2': FAIL,
};
```

and then 

    if (score === PASS)

but that is up to you:  The string constants are limited to this file, so do not require boilerplate yet.

src/playback/summary.jsx
```diff
@@ -226,16 +252,22 @@ class PlaybackSummary extends React.Component {
...
+                        <Tooltip
+                          content={this.tooltipContent(
+                            score,
+                            encoding,
+                            platform.id,
+                            id
+                          )}>
                           {label}
-                        </div>
+                        </Tooltip>
```
klahnakoski: move tooltip out so that it shows when hovering over the colored square, not just the text in the square.

klahnakoski: can the tooltip be wider?

Issue #532, opened 2019-08-12
Change power usage subtest names


gmierz:
A patch landed recently which changed the name of the subtests and they are no longer being displayed on the health graphics dashboard. There should be one test found in this range but it's not there: https://health.graphics/power?browser=geckoview&suite=speedometer&past=week&ending=2019-08-12

Here's a link to the change in subtest names: https://treeherder.mozilla.org/perf.html#/comparesubtest?originalProject=mozilla-central&newProject=mozilla-central&originalRevision=6a96fdc69cc1f05f318bb212d402142c73dd4026&newRevision=3afb892abb74c6d281f3e66431408cbb2e16b8c4&originalSignature=2027286&newSignature=2027286&framework=10

The following names were changed (the prefixed test name was removed 'raptor-scn-power-idle-geckoview-'):
`raptor-scn-power-idle-geckoview-power-cpu --> power-cpu`
`raptor-scn-power-idle-geckoview-power-wifi --> power-wifi`
`raptor-scn-power-idle-geckoview-power-screen --> power-screen`

If they could be pattern matched based on the subtest name ending (cpu, wifi, screen) that would be ideal in case any other changes occur.

PR #540, opened 2019-08-30
Issue 532 - Change power usage subtest names

airimovici:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/532

### DIFF
--- a/src/power/config.js
+++ b/src/power/config.js
@@ -6,7 +6,7 @@ const TESTS = [
   {
     id: 'cpu',
     label: 'CPU',
-    filter: { eq: { test: 'power-cpu' } },
+    filter: { in: { test: ['cpu', 'power-cpu'] } },
   },
   // {
   //   id: 'proportional',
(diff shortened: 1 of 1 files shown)


src/power/config.js
```diff
@@ -6,7 +6,7 @@ const TESTS = [
   {
     id: 'cpu',
     label: 'CPU',
-    filter: { eq: { test: 'cpu' } },
+    filter: { eq: { test: 'power-cpu' } },
```
davehunt: Will this also show the historical data? Perhaps this could be an 'or', or a suffix match?

klahnakoski: @davehunt is correct

```
{"or":[
    { eq: { test: 'cpu' } ,
    { eq: { test: 'power-cpu' } }
]}
```

or

```
{ in: { test: ['cpu', 'power-cpu' } }
```

airimovici: Thanks for pointing that out :)
I changed it to show historical data also:
health-power-historical

klahnakoski:
looks good

Issue #536, opened 2019-08-29
Change media playback results to Firefox Preview


davehunt:
Switch the media playback grid results on the Android summary page (https://health.graphics/android) to show results for Firefox Preview (Fenix) instead of GeckoView example.

Issue #531, opened 2019-08-09
Create media playback results page for Firefox Preview


davehunt:
We have started running the media playback tests on Firefox Preview. Let's add an option to select Firefox Preview from the media playback details page at https://health.graphics/playback

davehunt:
@airimovici could you have a look at this?

airimovici:
@davehunt Sure, there will be a PR for this issue and also https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/536

PR #541, opened 2019-09-02
Issues 536 + 531 - Change media playback results to fenix

airimovici:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/536 - Change media playback results to Firefox Preview
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/531 - Create media playback results page for Firefox Preview

### DIFF
--- a/src/playback/config.js
+++ b/src/playback/config.js
@@ -124,7 +124,7 @@ const PLATFORMS = [
 const BROWSERS =[
   {id: "firefox", label:"Firefox", filter: {eq: {framework: 10, repo: 'mozilla-central', suite: ["raptor-youtube-playback-firefox-live"]}}},
   {id: "geckoview", label:"geckoview", filter: {eq: {framework: 10, repo: 'mozilla-central', suite: ["raptor-youtube-playback-geckoview-live"]}}},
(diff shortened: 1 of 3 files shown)


src/playback/config.js
```diff
@@ -124,6 +124,7 @@ const PLATFORMS = [
 const BROWSERS =[
   {id: "firefox", label:"Firefox", filter: {eq: {framework: 10, repo: 'mozilla-central', suite: ["raptor-youtube-playback-firefox-live"]}}},
   {id: "geckoview", label:"geckoview", filter: {eq: {framework: 10, repo: 'mozilla-central', suite: ["raptor-youtube-playback-geckoview-live"]}}},
+  {id: "fenix", label:"fenix", filter: {eq: {framework: 10, repo: 'fenix', suite: ["raptor-youtube-playback-fenix-live"]}}},
```
davehunt: This should be labeled as "Firefox Preview". Also note in the deploy preview that there are several graphs with no data or data older than 3 days. I think we should increase the no data warning to a week on fenix.

airimovici: @davehunt So should we increase the missing data interval to 7 only on fenix and let firefox and geckoview have the default of 3?

davehunt: Yes, we expect Firefox and GeckoView to have multiple results a day. Fenix runs once/day but there may be no new commits for several days, so the results will appear to be older.

PR #545, opened 2019-09-09
Fix speedometer power usage graph

airimovici:
Replaced hardcoded test filter.
It works with historical data also.


klahnakoski:
excellent!

PR #546, opened 2019-09-09
Minor Travis change

armenzg:


klahnakoski:
@armenzg it looks like yarn will require an upgrade too

PR #548, opened 2019-09-11
Use fennec64 as reference

klahnakoski:
This PR is to pull fennec64 test results, and use them as a reference range


klahnakoski:
@arnold-iakab @airimovici    This is incomplete, but it does pull the reference data and calculates a range.  

The meaning of `PLATFORMS` was referring to the browser/os combination, but that has changed to `BROWSER_PLATFORMS`, which separates the the browser and the os.  This means we must be explicit with the `browser`; and you see it added in some calls.

src/android/TP6mAggregate.jsx
```diff
@@ -15,27 +17,108 @@ import { getData } from '../vendor/perfherder';
...
+  const referencePlatforms = selectFrom(BROWSER_PLATFORMS).where({ platform, browser: REFERENCE_BROWSER });
+  const referenceFilter = TP6_COMBOS
+    .where({
+      browser: REFERENCE_BROWSER,
+      platform,
+      test,
+    })
+    .select('filter')
+    .toArray();
+  const [fennec64, sources] = await Promise.all([
```
klahnakoski: @arnold-iakab @airimovici   after constructing the `referenceFilter`, we request the reference data along with all the other data

src/android/TP6mAggregate.jsx
```diff
@@ -15,27 +17,108 @@ import { getData } from '../vendor/perfherder';
...
+    getData({ or: referenceFilter }),
+    getData(condition),
+  ]);
 
   readData.done();
 
   const processData = timer('process data');
+
+  const reference = selectFrom(fennec64)
```
klahnakoski: @arnold-iakab @airimovici    Notice this block is near-identical to `measured = ...` directly below: Both the reference data and the measured data is processed the same.  This duplication is a bit unfortunate; if it happens a third time, then it belongs in a function.

src/android/TP6mAggregate.jsx
```diff
@@ -15,27 +17,108 @@ import { getData } from '../vendor/perfherder';
...
+        },
+      },
+    ]);
+  const refMax = window(
+    { reference },
+    {
+      edges: ['test', 'site', 'platform'],
+      value: ({ reference }) => selectFrom(reference).select('value').max(),
+    },
+  );
```
klahnakoski: These `window` functions are not well documented yet:  They operate on multidimensional pivot tables (aka cubes), each dimension/edge of the cube is given a name, the `edges` are the names we will iterate over; in this case over every test, site, platform combination; which which is all the dimensions.  This leaves {reference}, which is an array of data from perfherder, and we take the max of that.  

I suggest you put a debug stop point to better see what's happening in the `value` function. But first, since browser debuggers have been very terrible these past several years, you must expand the `value` function enough to add a breakpoint:

```
value: ({ reference }) => {
    const temp = selectFrom(reference).select('value').max();
    return temp;
},
```

src/android/TP6mAggregate.jsx
```diff
@@ -114,14 +197,14 @@ async function pullAggregate({
...
+    { daily, refMin },
     {
       edges: ['test', 'platform', 'pushDate'],
       value: (row) => {
-        const { daily, g5Reference } = row;
+        const { daily, refMin } = row;
 
         return round(
-          selectFrom(daily, g5Reference)
+          selectFrom(daily, refMin)
```
klahnakoski: now we can use `refMin` instead of the original g5Reference data.  In this case we are ensuring we have both reference data and sample data.

src/quantum/config.js
```diff
@@ -208,7 +208,7 @@ const PLATFORMS = [
   {
     browser: 'geckoview',
     label: 'Geckoview p2 aarch64',
-    platform: 'geckoview-p2-aarch64',
+    platform: 'p2-aarch64',
```
klahnakoski: @arnold-iakab @airimovici  and here we see clarifying the distinction between `platform`  and `browser`.  It should have been done a long time ago, but it was expedient at the time.  Now we need clarity so our reference data can match the `platform` of sample `platform`

klahnakoski:
@davehunt   Please do a visual review of "Raptor (TP6m)" section, and the breakdown view:

https://deploy-preview-548--firefox-health-dashboard.netlify.com/android
https://deploy-preview-548--firefox-health-dashboard.netlify.com/android/tp6m

The blue region shows the min and max values we are getting from fennec64

klahnakoski:
@davehunt yes, that is the plan.  If the min/max range is good, then the next PR is to link a full description like in the playback PR

PR #549, opened 2019-09-11
Filter out chrome release data from chromium data.

gmierz:
@klahnakoski here's what I've come up with so far for filtering real-chrome data out of chromium data based on time, and filtering chromium-named-chrome from real-chrome data. Not sure if this is the right spot to do this filtering or if this should happen elsewhere.


klahnakoski:
Sorry, the filtering should not be done here

Issue #251, opened 2019-01-07
Benchmarks with subtests to use PerfherderGraphContainer


armenzg:
For instance MotionMark, here's the code:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/quantum/index-64bit.js#L61-L72
Config code:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/quantum/config.js#L6-L35

Replacing the use of the `Perfherder` component will allow us removing the `Perfherder.js` component and take a step further towards moving to ChartJs as a plotting library.

No need to keep the "Talos: No regressions" line.

We need to keep to have the ability to take a user to a breakdown (this is a hyperlink):
<img width="67" alt="image" src="https://user-images.githubusercontent.com/44410/50773087-c3569280-125d-11e9-85e4-ecdae914eef2.png">
https://health.graphics/quantum/windows10-64/motionmark_htmlsuite

<img width="295" alt="image" src="https://user-images.githubusercontent.com/44410/50773011-8ab6b900-125d-11e9-9fdd-91d8f8cd5857.png">

armenzg:
See https://github.com/mozilla/firefox-health-backend/issues/53

PR #560, opened 2019-09-17
Replace benchmark drilldown view

klahnakoski:
The old D3 benchmark charts had dillldowns.  https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/251  This replaces them.


klahnakoski:
@davehunt  for visual review:

Added `...` for drilldown https://deploy-preview-560--firefox-health-dashboard.netlify.com/quantum/64

image

an example detail page: https://deploy-preview-560--firefox-health-dashboard.netlify.com/quantum/subtests?suite=motionmark-htmlsuite&platform=windows10-64

One problem is the Speedometer, which has many subttests, so it takes some time to load, and locks up for a bit too.

src/utils/PerfherderGraphContainer.jsx
```diff
@@ -323,18 +323,21 @@ class PerfherderGraphContainer extends React.Component {
...
             <div>
               {title}
-              {jointUrl && (
+              {moreUrl
+               && <a href={moreUrl}><DetailsIcon /></a>
+              }
+              {perfherderUrl && (
               <a
-                href={jointUrl}
+                href={perfherderUrl}
```
klahnakoski: Note the renaming of variable so it is not confused with the `moreUrl`

src/vendor/perfherder.js
```diff
@@ -145,33 +153,64 @@ function simpler(v) {
...
-    }
-
-    const lookup = Data.zip(props.map((p, i) => ([p, i])));
-    const inProps = array(props.length + 1).map(() => ([]));
-    if (isString(param)) {
-      const i = coalesce(lookup[param], props.length);
-      inProps[i].push({ [op]: param });
-      return [inProps.map(simpler)];
-    }
+const extract = (expression, props) => {
```
klahnakoski: The extraction function had to change to handle more complicated expressions

src/vendor/perfherder.js
```diff
@@ -256,16 +295,17 @@ const getData = async (condition) => {
 
   const results = await Promise.all(collated
     .map(async ([pushDate, repo, framework, rest]) => {
-      if (repo === true || framework === true) {
+      if (missing(repo) || missing(framework)) {
```
klahnakoski: The `extract` function now returns `null` for a missing expression rather than `true` because `null` is easier to filter out, and it makes more sense.  Plus, they both mean the same thing:  `true` as a filter is always true, `null` as a filter means no expression, which means no filtering, which is the same as using `true`.

davehunt:
I notice that were not showing the -shippable build variants, or chromium. Here's an example of what we should be showing: https://treeherder.mozilla.org/perf.html#/graphs?timerange=2592000&series=mozilla-central,1707657,1,10&series=mozilla-central,1730980,1,10&series=mozilla-central,1980705,1,10&series=mozilla-central,1939101,1,10&highlightAlerts=1

klahnakoski:
@davehunt  here is another version for visual review.  This time I also included the linux64

davehunt:
@klahnakoski looks good, could we also add macos? Example data https://treeherder.mozilla.org/perf.html#/graphs?timerange=2592000&series=mozilla-central,2054585,1,10&series=mozilla-central,2054495,1,10&highlightAlerts=1

klahnakoski:
maxosx added

PR #562, opened 2019-09-17
add playback summary to detail page

klahnakoski:


klahnakoski:
@davehunt for visual review.  I am not sure i like this.  You have other ideas?  Maybe increase the number of squares, for all the resolutions and refresh rates?  Instead of squares, maybe we color-code the title of the actual charts?  Maybe add a color border to any chart that shows we miss requirements?

https://deploy-preview-562--firefox-health-dashboard.netlify.com/playback?platform=mac&browser=firefox&encoding=VP9

image

airimovici:
If we will use it this way, I think we should show the squares grayed out for the cases where we don't have data.
It will be grayed out instead of red for the cases where the platform and the browser are incompatible (ex: macOSX - fenix), but also for the cases where we don't have data to show (>1080p tests on g5)

gray-for-no-data

@klahnakoski  Is this a change we would want?

klahnakoski:
@airimovici   I like the idea of showing missing data gray; it should be used on the summary page for sure.

For this detail page, I was more wondering about the number of squares showing:  Should they match the total number of resolutions tested?  Should we even show these squares, since the chart is already showing (albeit you may need to scroll).

klahnakoski:
@airimovici   ...maybe we show just the squares, and when we click on them, the corresponding chart shows?

davehunt:
I like the idea of adding a visual cue to each graph to show if they meet the criteria. I also think I like the idea increasing the squares in the at-a-glance blocks at the top to show overall health without needing to scroll.

Issue #572, opened 2019-10-01
Add Chrome results for all benchmarks


davehunt:
Now that we're running all tests against Chrome we should plot these in the benchmarks, such as Speedometer: https://health.graphics/quantum/subtests?suite=Speedometer&platform=win64&past=month

PR #573, opened 2019-10-01
enable chrome

klahnakoski:


klahnakoski:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/572

davehunt:
Preview looks good, but could we keep the colour of Chrome/Chromium consistent on the desktop summary? Currently page load has blue for Chromium and green for Chrome, but the benchmarks have orange/yellow for Chromium and blue for Chrome.

klahnakoski:
@airimovici The color is controlled by the order of the series given to the charting library.  There are three possible solutions (from easiest to hardest):

1. Reorder the browsers - the order may be preserved all the way through the processing pipeline, and everything "just works"
2. Add and `ordering` property to the `BENCHMARK_SUITES` objects, and then use that ordering just before we render, around https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/03047866ca815b3b34b58dc99e91969c18b4520a/src/quantum/Quantum.jsx#L96  insert a sort: `browsers.sort("ordering").map(....`
3. Update the charting library wrapper to accept a colour argument for a series; add a `color` property to a variety of browser configs, and use it at `render()` time.  See [schema](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/vendor/components/chartJs/chartSchema.md#series-property), and notice the property should be named `series.style.color`, and (ensure `cjsGenerator` deals with it properly](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/03047866ca815b3b34b58dc99e91969c18b4520a/src/vendor/components/chartJs/utils.js#L156..L162)

klahnakoski:
@airimovici since my change was so small, you can make your own PR that gets the colours right

PR #574, opened 2019-10-02
Feature/code splitting

the6thm0nth:
1. I updated to use `lazy`, `Suspense` for code-splitting only. and here are bundles.
```
                                       Asset       Size    Chunks                                Chunk Names
                        assets/0.83621cbb.js   33.4 KiB         0  [emitted] [immutable]         
                        assets/1.15984379.js   21.6 KiB         1  [emitted] [immutable]         
                       assets/19.6af1661a.js    658 KiB        19  [emitted] [immutable]  [big]  
                      assets/19.ee9c6117.css   4.26 KiB        19  [emitted] [immutable]         
                        assets/2.2c23cf71.js   12.1 KiB         2  [emitted] [immutable]         
                       assets/20.ac71f5a6.js   28.7 KiB        20  [emitted] [immutable]         
                      assets/21.a48568a9.css   7.43 KiB        21  [emitted] [immutable]         
                       assets/21.badd927c.js    104 KiB        21  [emitted] [immutable]         
                        assets/3.a75b2e58.js   12.1 KiB         3  [emitted] [immutable]         
                        assets/4.315af410.js   37.2 KiB         4  [emitted] [immutable]         
                        assets/5.2cb9b32f.js    534 KiB         5  [emitted] [immutable]  [big]  
```

some bundle files are still bigger than the standard size. I'd like to edit `webpack.config.js`, but it use `neutrino` that I'm not familiar with. is it ok to edit this file like this way 👉🏼?

```javascript
const neutrino = require('neutrino');

const config = neutrino().webpack();

module.exports = {
  ...config,
  blahblah: {
  }
}
```

2. I don't know why `lint` fixed some files I've not touched. I had to update snapshot because of this(maybe). did I by correct way?


klahnakoski:
> I'd like to edit webpack.config.js, but it use neutrino that I'm not familiar with. is it ok to edit this file like this way 👉🏼?

I do not familiar with neutrino either.  If you can get it to work, then it is good enough for me.

This project is a static web page: All that matters is the compiled bundle(s) work in the browser.  If your build/deploy infrastructure changes; it is not a big deal.

> I don't know why lint fixed some files I've not touched. I had to update snapshot because of this(maybe). did I by correct way?

The linter was updated a few months ago, and is more belligerent than ever.  If linting sneaks into your PR I will look past it.  I may ask some questions if it was you or not, just to be sure.

src/Home.jsx
```diff
@@ -44,27 +44,27 @@ const Home = ({ classes }) => (
     <div className={classes.links}>
       <Link to="/android">
         <AndroidIcon />
-        {' '}
+
```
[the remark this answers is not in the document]
the6thm0nth: by the linter 😢

src/routes.jsx
```diff
@@ -1,35 +1,38 @@
...
+const AndroidPage = lazy(() => import(/* webpackChunkName: "andriod" */'./android'));
+const NimbledroidGraphPage = lazy(() => import(/* webpackChunkName: "nimbledroid" */'./nimbledroid/NimbledroidGraphPage'));
+const Quantum = lazy(() => import(/* webpackChunkName: "quantum" */'./quantum/Quantum'));
+const TP6 = lazy(() => import(/* webpackChunkName: "quantum-tp6" */'./quantum/TP6'));
+const TP6M = lazy(() => import(/* webpackChunkName: "andriod-tp6m" */'./android/TP6m'));
+const Power = lazy(() => import(/* webpackChunkName: "power" */'./power'));
+const Playback = lazy(() => import(/* webpackChunkName: "playback-playback" */'./playback/playback'));
+const PlaybackDetails = lazy(() => import(/* webpackChunkName: "playback-details" */ './playback/details'));
+const Subtests = lazy(() => import(/* webpackChunkName: "quantum-subtests" */'./quantum/subtests'));
+const NoMatch = lazy(() => import(/* webpackChunkName: "not-found" */'./NotFound'));
```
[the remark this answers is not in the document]
the6thm0nth: I thought every page should be lazy loaded, but that's my misunderstading. I'll remove unnecessary chunks.

the6thm0nth:
1. I removed some chunks that seems unnecessary to lazy-load
2. I think it was not that major problem, but I tried to refactor not to use pure `metrics-graphics` library to draw a graphic. Instead, replaced with `react-metrics-graphics`

src/routes.jsx
```diff
@@ -1,35 +1,39 @@
...
 import PlaybackDetails from './playback/details';
-import Subtests from './quantum/subtests';
 
-const NoMatch = () => <div>404</div>;
+const AndroidPage = lazy(() => import(/* webpackChunkName: "andriod" */ './android'));
+const NimbledroidGraphPage = lazy(() => import(/* webpackChunkName: "nimbledroid" */ './nimbledroid/NimbledroidGraphPage'));
+const Quantum = lazy(() => import(/* webpackChunkName: "quantum" */ './quantum/Quantum'));
+const TP6 = lazy(() => import(/* webpackChunkName: "quantum-tp6" */ './quantum/TP6'));
+const TP6M = lazy(() => import(/* webpackChunkName: "andriod-tp6m" */ './android/TP6m'));
+const Subtests = lazy(() => import(/* webpackChunkName: "quantum-subtests" */ './quantum/subtests'));
```
klahnakoski: I still do not think we need a split by page.  We only need to split out large logic chunks, like the charting libraries.

src/telemetry/graph.jsx
```diff
@@ -23,79 +43,74 @@ class TelemetryContainer extends React.Component {
...
+      </Suspense>
+    );
   }
 
   render() {
     const { id, title } = this.props;
+    const { graphData } = this.state;
+
 
     if (title) {
```
klahnakoski: This is a big if block.  Consider 

    if (!title) return null;

src/telemetry/graph.jsx
```diff
@@ -1,15 +1,35 @@
...
-class TelemetryContainer extends React.Component {
+const MetricsGraphics = lazy(() => import(/* webpackChunkName: "react-metrics-graphics" */ 'react-metrics-graphics'));
+
+class TelemetryContainer extends Component {
+  state = {
+    graphData: null,
+    telemetryUrl: null,
+  };
+
+  get fullTelemetryUrl() {
```
klahnakoski: nice!

src/telemetry/graph.jsx
```diff
@@ -23,79 +43,74 @@ class TelemetryContainer extends React.Component {
...
-        path: telemetryUrl,
-        query: { processType: 'parent' },
+      this.setState({
+        graphData,
+        telemetryUrl,
       });
-
-      this.graphTitleLink.setAttribute('href', fullTelemetryUrl);
-      this.graphSubtitleEl.textContent = graphData.description;
-      this.graphEvolutionsTimeline(graphData, this.graphEl);
```
klahnakoski: nice cleanup!

src/telemetry/graph.jsx
```diff
@@ -23,79 +43,74 @@ class TelemetryContainer extends React.Component {
...
+    const {
+      graphData: {
+        datas, params, yLabel, legendLabels,
+      },
+    } = this.state;
+
+    return (
+      <Suspense fallback="">
+        <MetricsGraphics
+          aggregate_rollover
```
klahnakoski: Nice convert to jsx

src/telemetry/graph.jsx
klahnakoski: ...oh but it seems that's not your line.  Nevermind.

src/routes.jsx
klahnakoski: ...but this is not a blocker

the6thm0nth:
1. changed my code to split chunks by large components(👉🏼 `metrics-graphic`) not pages.
2. refactored a few lines by your advise 👍

the6thm0nth:
I guess `a is null` error is caused by replacing pure `metrics-graphic` to `react-metrics-graphic`. I'll find it out.

klahnakoski:
I am not too concerned with the `r is null` error, since a number of the charts never did show due to an error of some kind.  It is the missing charts that is the problem.

src/telemetry/graph.jsx
```diff
@@ -27,12 +26,12 @@ class TelemetryContainer extends Component {
     return null;
   }
 
-  async componentDidMount() {
-    const { id, queryParams } = this.props;
-    await this.fetchPlotGraph(id, queryParams);
+  componentDidMount() {
+    this.fetchPlotGraph();
```
klahnakoski: @the6thm0nth   Please keep this async:  
```
  async componentDidMount() {
    return this.fetchPlotGraph();
  }
```

or

```
  async componentDidMount() {
    await this.fetchPlotGraph();
  }
```

This project has higher level components that wrap `componentDidMount()` in a try/catch, just in case anything goes wrong.  Calling an async function like a function breaks this error handing

When you call an async function like a function (without await), you are spawning a new promise chain. Promise chains are like threads; they are sequences of control that run independently from the main flow of control.  You will and will loose any return values, or thrown exceptions from it; your code is now responsible for communicating between the main flow of control and the new promise chain.  Spawning new promise chains is powerful, so there are many good times to use them.  Here is the pattern:

    // define an async function
    const myChain = sync () => {
       /*code for a new promise chain*/
    }
    // run it now
    (myChain)();

sometimes written all together,  to save ourselves naming things:

    (async ()=>{ 
        /*code for a new promise chain*/  
    })(); 

So, whenever you see an async function called as a regular function, realize you are loosing some control.

the6thm0nth: Oh, sorry for that, I didn't know this is used by higher level components. I fixed it.

klahnakoski:
@the6thm0nth any more updates on this?  If not, I can hunt down the final problem and get it merged

the6thm0nth:
sure :) I was waiting for your answer. I would be my pleasure if it is merged.

klahnakoski:
@the6thm0nth sorry for yet-another delay.  Please @klahnakoski so I see the messages you post here.

What is the question you are waiting for?

klahnakoski:
Looking a bit more into why the charts are not showing 

https://stackoverflow.com/questions/31268920/using-metricsgraphics-js-with-reactjs

also, a line from the old code is missing

    const newDatas = datas.map(evo => MG.convert.date(evo, 'date'));

PR #586, opened 2019-10-14
fix: broken link to neutrino.js description

driver-by:
I found that Neutrino "here" link is [broken](https://github.com/neutrinojs/neutrino/blob/master/docs/packages/react/README.md#features). I've changed it to [working one](https://github.com/neutrinojs/neutrino/blob/master/packages/react/README.md#features)


klahnakoski:
Thank you very much!

Issue #597, opened 2019-11-11
Android tp6 summary graph for Pixel 2 links to results for Moto G5


davehunt:
The link from the Android summary page to view details of the tp6 results on Pixel 2 loads the results for Moto G5.

<img width="833" alt="Screenshot 2019-11-11 at 21 53 19" src="https://user-images.githubusercontent.com/122800/68624260-5a1f5b80-04ce-11ea-874d-e70a4ef1c58c.png">

<img width="549" alt="Screenshot 2019-11-11 at 21 58 45" src="https://user-images.githubusercontent.com/122800/68624369-8b982700-04ce-11ea-95e0-b622d9670e44.png">

PR #600, opened 2019-11-13
Issue 597 - Fix tp6m summary graph link to more details

airimovici:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/597

### DIFF
--- a/src/android/TP6mAggregate.jsx
+++ b/src/android/TP6mAggregate.jsx
@@ -325,10 +325,9 @@ class TP6mAggregate_ extends Component {
               const platform = row.platform.getValue();
               const count = row.count.getValue();
               const total = row.total.getValue();
-              const browserPlatform = selectFrom(BROWSER_PLATFORMS)
+              const { label: platformLabel, id: browserPlatform } = selectFrom(BROWSER_PLATFORMS)
                 .where({ browser, platform })
                 .first();
-              const platformLabel = browserPlatform.label;
 
               return (
                 <Grid item xs={6} key={platform}>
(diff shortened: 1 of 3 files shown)


airimovici:
Link for testing: 
https://deploy-preview-600--firefox-health-dashboard.netlify.com/android

src/android/TP6mAggregate.jsx
```diff
@@ -342,7 +342,7 @@ class TP6mAggregate_ extends Component {
                           path: '/android/tp6m',
                           query: {
                             test,
-                            platform,
+                            platform: browserPlatform.id,
```
klahnakoski: Writing `platform: browserPlatform.id` means something is not named correctly:  ES6 re/destructuring promotes using a single name across functions and modules for the same thing.  When you see a name change, either something has the wrong name, or logic is wrong.  I am concerned that `platform` is different from `browserPlatform.id`.   I am guessing the problem you are trying to solve is somewhere else in the code.

* is `platform` value correct?  maybe it must be changed to `"fenix-*"`?

klahnakoski: @airimovici  a gentle reminder that this PR is still open.

airimovici:
The problem here is that the nav expects a browser-platform id and currently it is given a browser-id. With this wrong value, it will always got ot the default value: 'fenix-g5'

What can be done to eliminate the confusion is to change the nav picker id from TP6m page from platform to browserPlatform.

src/android/TP6mAggregate.jsx
airimovici: There is a confusion between platform and browserPlatform in several components and parts of the code. For this case in particular I suggested a fix below: to eliminate the confusion we should change the nav picker id from TP6m page from `platform` to `browserPlatform`

airimovici:
@klahnakoski I did the changes as I described above

PR #618, opened 2019-12-11
Faster Travis results

armenzg:
Separate tests, coverage, linters and build step to get faster results.


armenzg:
This is only slightly faster (1min). Not sure why I had expected more.
image

Issue #570, opened 2019-09-24
Show correct links in graph tooltip for fenix


airimovici:
The tooltip links from PerfherderGraphContainer have `mozilla-central` as hardcoded repo. This does not work for fenix tests, which are on the `fenix` repo.

divyanshu-rawat:
@airimovici is this issue open to work on ?

airimovici:
@divyanshu-rawat it is available. If you want to work on it and have any questions I can help you with that.

divyanshu-rawat:
cool, 🎉 Happy to give back something to the community @airimovici , let me know what it is all about?

airimovici:
@divyanshu-rawat 
Basically, we have these graphs (PerfherderGraphContainer) in several pages and they show results from tests. When hovering on a point from one graph, there is a tooltip with info about that result that also contains some external links.

![graph-tooltip-rectangle](https://user-images.githubusercontent.com/51707553/71096422-f10ac200-21b6-11ea-9ac6-84ac8f221f74.png)

[Example of page with these type of graphs](https://health.graphics/playback/details?platform=win-aarch64&browser=firefox&encoding=VP9&past=week&ending=2019-12-18)

For a while we only had tests from a single repo: `mozilla-central`
The reason for the existence of this bug is that lately there are some tests that are from another repo (`fenix`), and the repo is hardcoded when creating the external links.

[Example of page where the external links in the tooltip are broken](https://health.graphics/playback/details?platform=p2-aarch64&browser=fenix&encoding=VP9&past=week&ending=2019-12-18)

PR #635, opened 2019-12-21
Fix for #570: Broken Links Fixed.

divyanshu-rawat:
Fix for #570 

@airimovici feel free to review 👍 🎉

### DIFF
--- a/src/utils/PerfherderGraphContainer.jsx
+++ b/src/utils/PerfherderGraphContainer.jsx
@@ -60,13 +60,15 @@ const tip = withStyles(tipStyles)(
     const higherOrLower = series.meta.lowerIsBetter
       ? 'lower is better'
       : 'higher is better';
-    const hgURL = URL({
-      path: `https://github.com/mozilla-mobile/fenix/commit/${record.revision}`,
-    });
+
+    const revisionURL = record.meta.repo === 'mozilla-central'
+      ? `https://hg.mozilla.org/mozilla-central/pushloghtml?changeset=${record.revision}`
+      : `https://github.com/mozilla-mobile/fenix/commit/${record.revision}`;
+
     const jobURL = URL({
       path: 'https://treeherder.mozilla.org/#/jobs',
       query: {
-        repo: 'fenix',
+        repo: record.meta.repo,
         revision: record.revision,
         selectedJob: record.job_id,
         group_state: 'expanded',
@@ -126,7 +128,7 @@ const tip = withStyles(tipStyles)(
           );
         })()}
         <div>
-          <a href={hgURL} target="_blank" rel="noopener noreferrer">
+          <a href={revisionURL} target="_blank" rel="noopener noreferrer">
             {record.revision.slice(0, 12)}
           </a>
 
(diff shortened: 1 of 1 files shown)


src/utils/PerfherderGraphContainer.jsx
```diff
@@ -61,13 +61,12 @@ const tip = withStyles(tipStyles)(
       ? 'lower is better'
       : 'higher is better';
     const hgURL = URL({
-      path: 'https://hg.mozilla.org/mozilla-central/pushloghtml',
-      query: { changeset: record.revision },
+      path: `https://github.com/mozilla-mobile/fenix/commit/${record.revision}`,
```
klahnakoski: Also, do not use string munging to make URLs:  Although the `record.revision` happens to have URL-safe characters, this is not always the case.

divyanshu-rawat: So, what other alternative one can have instead of munging, concatenation ?

divyanshu-rawat:
> * mozilla-central

So, is hgURL fine, do I have to work on `jobURL` now to fix it?
Though, I didn't understood this statement, 
`The objective is to put find out where repo value is coming from and use it to populate this URL.`

src/utils/PerfherderGraphContainer.jsx
klahnakoski: Undo this change.   

Ensure your programming environment has a keyboad shortcut that will **jump to definition**, where comments may describe the function you are using.  Ensure you have a keyboard shortcut to **find all usages** so you can see how other parts of the program use a class/function

klahnakoski:
#570 is about enhancing the tooltip so it can point to either  `mozilla-central` or `fenix`, depending on the sequence being shown.

The `repo: "fenix"` comes from here: https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/f3f1363b7264df08201ec52c24c9da03bd064f15/src/playback/config.js#L127

Using the debugger, at the point where where the tooltip gets created, we can see the many values being passed.  Some of those values come from the `config.js` file

image

divyanshu-rawat:
@klahnakoski will have a look and update the PR accordingly  👍

divyanshu-rawat:
@airimovici let me know if it is fine now :)

src/utils/PerfherderGraphContainer.jsx
```diff
@@ -60,14 +60,15 @@ const tip = withStyles(tipStyles)(
...
       ? 'lower is better'
       : 'higher is better';
-    const hgURL = URL({
-      path: 'https://hg.mozilla.org/mozilla-central/pushloghtml',
-      query: { changeset: record.revision },
-    });
+
+    const revisionURL = record.meta.repo === 'mozilla-central'
+      ? `https://hg.mozilla.org/mozilla-central/pushloghtml?changeset=${record.revision}`
+      : `https://github.com/mozilla-mobile/fenix/commit/${record.revision}`;
```
klahnakoski: This should not be text munging.  It would be nice if the URL was yet-another-property, instead of switching in "mozilla-central".

In any case, this is good enough.

divyanshu-rawat: Thanks @klahnakoski  👍😀

PR #641, opened 2020-01-03
Add descriptive page titles

### LINKED ISSUE #628: Add descriptive page titles
When attempting to navigate back without cycling through all pages (which is very slow) I'd like to be able to jump back to a previously visited page. The problem is, all pages have the same title, which makes this very difficult.

Screenshot 2019-12-12 at 22 06 04

Fcmam5:
Hi!
Thank you for all your efforts for keeping Mozilla alive! This is my first PR, this one should resolve #628

### DIFF
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -36,3 +36,14 @@ export const determineStatusColor = (data, key, target) => {
 
   return 'red';
 };
+
+// get a window title by concatenating the given title and the base title (Firefox Health Dashboard)
+// ignore if pageTitle is the same as baseTtitle
+export const getWindowTitle = pageTitle => {
+  const baseTitle = 'Firefox Health Dashboard';
+  return (
+    pageTitle && !baseTitle.toLowerCase().includes(pageTitle.toLowerCase())
+  )
+    ? `${pageTitle} - ${baseTitle}`
+    : baseTitle;
+};
--- a/test/utils/helpers.test.js
+++ b/test/utils/helpers.test.js
@@ -1,0 +1,22 @@
+/* eslint-disable jest/valid-expect */
+/* global describe, it */
+import { expect } from 'chai';
+import { getWindowTitle } from '../../src/utils/helpers';
+
+describe('helpers', () => {
+  const baseTitle = 'Firefox Health Dashboard';
+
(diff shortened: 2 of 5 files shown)


src/utils/helpers.js
```diff
@@ -36,3 +36,14 @@ export const determineStatusColor = (data, key, target) => {
...
+// get a window title by concatenating the given title and the base title (Firefox Health Dashboard)
+// ignore if pageTitle is the same as baseTtitle
+export const getWindowTitle = pageTitle => {
+  const baseTitle = 'Firefox Health Dashboard';
+  return (
+    pageTitle && !baseTitle.toLowerCase().includes(pageTitle.toLowerCase())
+  )
+    ? `${pageTitle} - ${baseTitle}`
+    : baseTitle;
+};
```
klahnakoski: Please inline this function, or make it a member function of `DashboardPage`.  At the very least, put the function in `DashboardPage.jsx` so it is "near" the only caller.

In general, we should not make names for things with only one call point: This code is specific to the caller's needs, it can not be used anywhere else; it is best to inline the block with suitable comments (like you provided).

test/utils/helpers.test.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable jest/valid-expect */
+/* global describe, it */
+import { expect } from 'chai';
+import { getWindowTitle } from '../../src/utils/helpers';
+
+describe('helpers', () => {
```
klahnakoski: Thank you very much for additional tests!

src/utils/helpers.js
Fcmam5: I moved it here just to make testing possible, I couldn't figure out how to get the component instance and then the member function to test it since the component is wrapped into `withStyles`.

Or should we just install jsdom and try to mock `document.title` ?

klahnakoski: Oh yes, I forgot about your tests on the function.  In which case, please ensure the function is in the `DashboardPage.jsx`, and leave it at that.

I assumed the template tests would have checked for the document title, but I now see they do not.  

In the future, small functions do not need automated tests: They can be manually tested when reviewed, and manually tested when changed.  This particular function has little risk:  The worst that can happen is a badly formatted title.

Issue #466, opened 2019-07-04
most charts not locking to timeDomain


klahnakoski:
if there are no datapoints at either extreme of the timeRange, the chart does not show the full time range, only the range of time with points

grantlouisherman:
hey @klahnakoski I can take at look at this issue. Could you provide a screen shot of what you are talking about?

klahnakoski:
You can see the problem here:

https://health.graphics/android/tp6m?test=cold-loadtime&browserPlatform=fenix-g5&past=month&ending=2019-11-19


The chart should be showing a full month ending November 19, which means the first date should be October19.  The chart starts on October26.

![image](https://user-images.githubusercontent.com/2334429/69195066-c9153980-0af8-11ea-92e0-b160297a5e50.png)

KaairaGupta:
hey @klahnakoski ! is this issue still up? I would like to work on it

KaairaGupta:
@klahnakoski I have set up my environment. Since this is my first contribution can you guide me as to how to proceed further?

grantlouisherman:
Hey @KaairaGupta you can take it on. I figured out the bug is coming from how the program creates dates. I believe there is a date component you can look at.

KaairaGupta:
@grantlouisherman can you please guide me as to which files i should look at?

klahnakoski:
The problem can be seen here: In the `moreProps.options.scales.xAxis`, which is sent to the ChartJs library.  The code responsible for setting this value is busted.

![image](https://user-images.githubusercontent.com/2334429/71547452-2610d480-296e-11ea-925e-9ebb8224aab0.png)

Looking for `xAxes` will reveal the code:  And we now have a good hint about why the xAxes is not set properly.

![image](https://user-images.githubusercontent.com/2334429/71547514-d8489c00-296e-11ea-9790-083175a9eed4.png)


Please notice:  I am using the Firefox debugger, and I have the React debug tools installed.

vaideshshank:
@klahnakoski if the issue has not been resolved yet, then can I take it up ? I have been looking upon this issue for a while now and would be happy to resolve it. @KaairaGupta if you are done with solving the issue, please let me know.

KaairaGupta:
@vaideshshank if you have made any progress with it, you can work on it

klahnakoski:
@vaideshshank This issue is still not resolved.  Please go ahead and work on it.

vaideshshank:
[...]

PR #642, opened 2020-01-06
charts locked to time domains

vaideshshank:
Here I have resolve issue #466 .
The charts have been locked to time domain by passing the labels as prop to the Chart component. Data points get distributed within the time range passed.


klahnakoski:
Rather than generating the labels, which will depend on the size of the range, and depend on if the min and max are even provided:  Ensure ChartJS receives the min/max; it can handle the variety of edge cases.

vaideshshank:
Okay that's a nice suggestion. Will make the changes soon.

Issue #626, opened 2019-12-11
Media Playback - Criterion change


klahnakoski:
The media plackback should color according to 

* greater than 1 dropped frame for 1080p 30fps (or smaller, or slower) - RED 
* greater than 1 dropped from on all others - YELLOW

airimovici:
@klahnakoski I'm not sure if I understand this issue correctly and I have some questions:
1. By slower, do you mean speeds below 1x playback speed?
2. I'm not sure sure what are the differences between the current criterion and the one we should change to. I will write what I understand it changes, please point out if I am missing an edge case.
From my understanding, the differences are:
	- 1080p 60fps, 1440p and 2160p can be at maximum yellow (now it can also be red)
	- for the example in the attached image, it should be red according to the new criteria.
![test-example](https://user-images.githubusercontent.com/51707553/71913843-01dca100-3181-11ea-9345-008d9b34121e.png)

klahnakoski:
@airimovici  The old criterion only cared about 1x playback speed.  The new criterion will go red on slower playback too.  The resolution is now in the criterion too: before high res failure was marked red, now it can only be marked yellow.  Here is the old logic.   

          lookupType[
            selectFrom(speeds)
              .map(({ speed, loss }) => {
                if (speed === 1) {
                  if (missing(loss)) return 3;
                  if (loss > 1) return 2;

                  return 0;
                }

                if (missing(loss) || loss <= 1) return 0;

                return 1;
              })
              .max()
          ],

PR #645, opened 2020-01-15
Issue 626 - Media playback criterion change

airimovici:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/626


airimovici:
I also added some visual changes in this PR:
- fixed the loading spinner not to be overlapped with the next component
- added spacing for the titles, not to overlap with the results anymore

airimovici:
@klahnakoski I didn't change the special sizes we had defined:

```
const SPECIAL_SIZES = [
  { id: '480p30', label: '480p' }, // .4M pixels
  { id: '720p60', label: '720p' }, // 1M pixels
  { id: '1080p60', label: '1080p' }, // 2M pixels
  { id: '1440p60', label: '1440p' }, // 4M pixels
  { id: '2160p60', label: '2160p' }, // 8M pixels
```
But in the issue description there is this:

> The media plackback should color according to:
> - greater than 1 dropped frame for 1080p **30fps** (or smaller, or slower) - RED
> - greater than 1 dropped from on all others - YELLOW

Did you mean there 1080p 60fps? The criterion change I did in this PR considers the first 3 sizes (480p30, 720p60, 1080p60) from the above SPECIAL_SIZES as "small sizes" and colors them red according to the criterion above.

My question is: Should 1080p60 considered under the line in the criterion and can be colored RED also, or is it over the line and can be colored YELLOW also?

klahnakoski:
I was specifically not meaning  1080p 60fps.,it is too fast.  I *think* all 60fps will be considered fast, no matter how small the resolution.

airimovici:
In this case we need to change the SPECIAL_SIZES definition to include only results for 30fps on the tests we are interested in:

```
const SPECIAL_SIZES = [
  { id: '480p30', label: '480p' }, // .4M pixels
  { id: '720p30', label: '720p' }, // 1M pixels
  { id: '1080p30', label: '1080p' }, // 2M pixels
  { id: '1440p30', label: '1440p' }, // 4M pixels
  { id: '2160p30', label: '2160p' }, // 8M pixels
]
```

src/playback/summary.jsx
```diff
@@ -140,30 +140,36 @@ class PlaybackSummary extends React.Component {
...
-
-                  return 0;
-                }
-
-                if (missing(loss) || loss <= 1) return 0;
-
-                return 1;
-              })
+              .map(({ size, speed, loss }) => this.getPlaybackScore(
+                size, speed, loss, sizes.slice(0, 3),
```
klahnakoski: `sizes.slice(0, 3),` is magic.  Please update the config so that you do not need special code like this.

src/playback/summary.jsx
```diff
@@ -140,30 +140,36 @@ class PlaybackSummary extends React.Component {
...
       .toArray();
-
     this.setState({ scores: result });
   }
 
   async componentDidMount() {
     await this.update();
   }
 
+  getPlaybackScore(size, speed, loss, smallSizes) {
```
klahnakoski: good pulling this out.  but remove the smallSizes, and use the CONFIG directly

airimovici:
I changed the config to have the data if a size is considered small and a speed is considered slow and now we get them like that.
```
const SMALL_SIZES = selectFrom(SPECIAL_SIZES).where({ isSmall: true }).select('id').toArray();
const SLOW_SPEEDS = selectFrom(SPEEDS).where({ isSlow: true }).select('speed').toArray();
```

src/playback/summary.jsx
```diff
@@ -140,30 +144,36 @@ class PlaybackSummary extends React.Component {
...
+
+    if (SLOW_SPEEDS.includes(speed)) {
+      if (SMALL_SIZES.includes(size) && loss > 1) {
+        return 2;
+      }
+      if (loss > 1) return 1;
+      return 0;
+    }
+
+    if (missing(loss) || loss <= 1) return 0;
```
klahnakoski: move this up so you can remove the `loss > 1` checks

klahnakoski:
```
  getPlaybackScore(size, speed, loss) {
    if (speed === 1 && missing(loss)) return 3;
    if (missing(loss) || loss <= 1) return 0;

    if (SLOW_SPEEDS.includes(speed) && SMALL_SIZES.includes(size)) return 2;
    return 1;
  }
```

src/playback/summary.jsx
```diff
@@ -172,7 +182,7 @@ class PlaybackSummary extends React.Component {
       .where({ encoding, platform, size })
       .first();
 
-    if (score === 'bad') {
+    if (score === 'bad' || score === 'fail') {
       result.speeds.sort((a, b) => a.speed - b.speed);
 
       const speed = result.speeds.find(s => s.loss > 1);
```
klahnakoski: I am not sure which direction you are sorting; I assume you are looking for the lowest speed.  Do this with one expression:

```
selectFrom(result.speeds)
    .sort("speed")
    .filter(s=>s.loss>1)
    .first()
```

Issue #653, opened 2020-01-24
Show "Windows 64" in title 


klahnakoski:
https://health.graphics/windows/64  looks like 

![image](https://user-images.githubusercontent.com/2334429/73087844-22a33700-3ea1-11ea-9fec-908582f34dbd.png)

have it say "Windows 64".  Do the same for the 32 bit version

the-pro:
hello sir, i want to take this issue if it is still available.

klahnakoski:
@the-pro  yes please.  it's yours now.

the-pro:
So i tried to make the change. Tell me if this was what you needed:
![windows](https://user-images.githubusercontent.com/42973786/73136168-b5de9700-4070-11ea-9077-d1b7a772e2a1.png)

klahnakoski:
@the-pro  Yes, that's what i am looking for.  Please make a PR

the-pro:
Done.

rimshach:
Hey , I read the comments and it seems like issue has been resolved however if it is still open I want to take this issue

klahnakoski:
@rimshach yes. this is resolved

PR #654, opened 2020-01-26
Update Windows.jsx

the-pro:
Wrote a patch of code to change the title as mentioned in (#653 ).

### DIFF
--- a/src/windows/Windows.jsx
+++ b/src/windows/Windows.jsx
@@ -659,12 +659,20 @@ export default class WindowsIndex extends Component {
         </DashboardPage>
       );
     }
-
-    return (
-      <DashboardPage title="Windows" subtitle="Release Criteria Report">
-        {reduced}
-      </DashboardPage>
-    );
+    if (bits === 32) {
+      return (
+        <DashboardPage title="Windows 32" subtitle="Release Criteria Report">
+          {reduced}
+        </DashboardPage>
+      );
+    }
+    if (bits === 64) {
+      return (
+        <DashboardPage title="Windows 64" subtitle="Release Criteria Report">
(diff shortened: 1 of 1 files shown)


src/windows/Windows.jsx
```diff
@@ -659,12 +659,20 @@ export default class WindowsIndex extends Component {
...
+        </DashboardPage>
+      );
+    }
+    if (bits === 64) {
+      return (
+        <DashboardPage title="Windows 64" subtitle="Release Criteria Report">
+          {reduced}
+        </DashboardPage>
+      );
+    }
```
klahnakoski: Use the `bits` variable right tin the template, do not make `if/else` (or `if/if` in your case) blocks:

Example

    <DashboardPage title={"Windows "+ bits} ....

klahnakoski (changes requested):

src/windows/Windows.jsx
the-pro: Done sir. Thanks for the review.

klahnakoski (changes requested):
Almost done.

1. Run `yarn lint --fix`, and fix any errors it complains about
2. Run `yarn test -u` to update the test templates (which will capture the 64/32 changes)
3. Commit-and-push those changes
4. Ensure all the travis checks pass

klahnakoski (changes requested):
@the-pro   This still does not pass tests; I do not know how the checks managed to miss that. Be sure to stage your changes then

    git commit -m "your message... 
    git push origin the-pro:patch-1 


image

klahnakoski:
Sorry for the delay merging this.

Issue #639, opened 2019-12-23
Performance issue on media playback summary page


airimovici:
This slowdown issue was raised here: https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/632#pullrequestreview-334073165

klahnakoski:
@airimovici Are you working on this?

airimovici:
@klahnakoski I did some tests on it and managed to drop it down from 1:30-1:50 min to 30-35 sec with limiting the scanning of signatures.

airimovici:
What I did was only scanning on the video playback signatures instead of all of them from perfherder. This reduced the signatures to scan on from ~60k to ~8k.

klahnakoski:
@airimovici is there a PR for this?

airimovici:
I created this PR: https://github.com/mozilla-frontend-infra/firefox-health-dashboard/pull/658

PR #658, opened 2020-01-27
Issue 639 - Improve perfomance for playback summary

airimovici:
Suggested change for improving the performance on the playback summary.

This should reduce the page loading time to almost half of the current value.

### DIFF
--- a/src/vendor/perfherder.js
+++ b/src/vendor/perfherder.js
@@ -24,7 +24,11 @@ const MIN_CHUNK_SIZE = 10; // ACCUMULATE SIGNATURE REQUESTS TO A SINGLE REQUEST
 const PERFHERDER = {
   signatures: [],
 };
+const PLAYBACK = [];
 const TREEHERDER = 'https://treeherder.mozilla.org';
+const TEST_TYPE = {
+  playback: 'playback',
+};
 const getAllOptions = (async () => {
   const output = await fetchJson(
     URL({ path: [TREEHERDER, 'api/optioncollectionhash/'] }),
@@ -66,10 +70,12 @@ const getFramework = async ({ repo, framework }) => {
           })();
           let lowerIsBetter = lower_is_better;
           let unit = 'Score';
+          let isPlayback = false;
 
           if (suite.includes('youtube-playback')) {
             lowerIsBetter = true;
             unit = 'count';
+            isPlayback = true;
 
             if (exists(test) && test.includes('%')) {
               unit = '';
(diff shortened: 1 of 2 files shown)


airimovici:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/639

airimovici:
For comparing the loading times:

https://health.graphics/playback vs. https://deploy-preview-658--firefox-health-dashboard.netlify.com/playback

src/vendor/perfherder.js
```diff
@@ -126,11 +132,13 @@ const getFramework = async ({ repo, framework }) => {
             id: meta.id,
             framework: meta.framework_id,
             repo,
+            isPlayback,
           };
         })
         .toArray();
 
       PERFHERDER.signatures.push(...clean);
+      PLAYBACK.push(...(clean.filter(({ isPlayback }) => isPlayback)));
```
klahnakoski: This is very specific.  I see two logical paths weaved through the code; the original and the `isPlayback` paths.  The `isPlayback` subtitly depends on the original path; preventing us from improving the original; like making it faster in general. If we keep this code, then we must add tests for the `isPlayback` path so we do not beak it accidentally.  The original path did not need tests because it is used everywhere; its failure would be obvious on any page. If we move away from one logical path, we will need automated testing on those paths because a human will forget to test them all manually. 

Your specific solution does allude to a general solution: Filter the main list to make a short-list of tests.  Since the playback tests use a type of `suite` filtering (eg `suite: ["raptor-youtube-playback-fenix-live"]`), we can take advantage of that:

We can split a filter using the (badly named) `extract` function:

```
/*
 * Expecting props to be an array of N property names
 * Return a list of expression tuples; meant to represent disjunctive normal form
 * such that
 *   output[x][i] contains expression on props[i]
 *   output[x][N] is expressions on everything else
 */
const extract = (expression, props) => {
```

and use it like so

    suite_filter, residue_filter = extract(filter, ["suite"]);

we can make a general cache holding the complete list of signatures for every suite

    cache = {}

if the suite_filter looks like 

    suite_filter == {eq: {suite: suite_name}}
    suite_filter == {eq: {suite: ["raptor-youtube-playback-fenix-live"]}}

then we can jump to a the specific signature list:

    short_list = cache[suite_name]

if `cache[suite_name]` is missing, we can build it.  If there is no clean suite filter (rare, but it happens) then we can simply use the full `PERFHERDER.signatures`.

Your code already identifies the places that need to change; your code should be enhanced to handle `suite` in general.

klahnakoski (changes requested):

src/vendor/perfherder.js
airimovici: Thanks for the detailed review. I thought about a more general approach on this (in the end I forced this simple approach only for video playback tests), but your idea is a lot better than what I though about. I haven't had the time this week to work on it

PR #662, opened 2020-02-12
Issue 333 - Improve Coverage

wicked-kiD:
It's my first PR here and I will be glad to see your comments.

### DIFF
--- a/src/vendor/convert.js
+++ b/src/vendor/convert.js
@@ -108,7 +108,9 @@ convert from bytes to base64 (encoded to be URL safe)
  */
 function bytesToBase64URL(bytes) {
   if (isString(bytes)) {
-    return null.f();
+    const type = typeof bytes;
+    // eslint-disable-next-line no-console
+    return console.error('expecting array of ints, not', type);
   }
   return window.btoa(String.fromCharCode(...new Uint8Array(bytes)))
     .replace(/[+/=]/g, m => toURL[m]);
--- a/src/vendor/convert.js
+++ b/src/vendor/convert.js
@@ -1,7 +1,7 @@
 import { length, selectFrom, toPairs } from './vectors';
 import { Log } from './logs';
 import { isData } from './datas';
-import { isArray, isFunction } from './utils';
+import { isArray, isFunction, isString } from './utils';
 import strings from './strings';
 
 function json2value(json) {
@@ -107,7 +107,10 @@ const toURL = { '+': '-', '/': '_', '=': '' };
 convert from bytes to base64 (encoded to be URL safe)
  */
 function bytesToBase64URL(bytes) {
-  return window.btoa(String.fromCharCode(...Array.from(new Uint8Array(bytes))))
+  if (isString(bytes)) {
+    return null.f();
+  }
+  return window.btoa(String.fromCharCode(...new Uint8Array(bytes)))
     .replace(/[+/=]/g, m => toURL[m]);
 }
 
--- a/test/vendor/convert.test.js
+++ b/test/vendor/convert.test.js
@@ -1,5 +1,11 @@
 /* global describe, it */
-import { json2value, value2json } from '../../src/vendor/convert';
+import {
+  json2value,
+  value2json,
+  escapeRegEx,
+  bytesToBase64URL,
+  base64URLToBytes,
+} from '../../src/vendor/convert';
 
 describe('convert', () => {
   it('value2json', () => {
@@ -36,5 +42,20 @@ describe('convert', () => {
     expect(() => json2value("'''")).toThrow();
     expect(json2value('"text"')).toEqual('text');
     expect(json2value('{"b": 2}')).toEqual({ b: 2 });
+  });
+
+  it('escapeRegEx', () => {
+    expect(escapeRegEx('q2W#e.[{]}')).toBe('q2W#e\\.\\[\\{\\]\\}');
+    expect(escapeRegEx('./:loiep*&')).toBe('\\./:loiep\\*&');
+  });
+
+  it('bytesToBase64URL', () => {
+    expect(bytesToBase64URL('0000001')).toBe('AA');
+    expect(bytesToBase64URL('0000002')).toBe('AAA');
+  });
+
+  it('base64URLToBytes', () => {
+    expect(base64URLToBytes('aHR0cHM6Ly9nb29nbGUuY29t')).toBe('https://google.com');
+    expect(base64URLToBytes('aHR0cHM6Ly93d3cubW96aWxsYS5vcmc=')).toBe('https://www.mozilla.org');
   });
 });
--- a/test/vendor/convert.test.js
+++ b/test/vendor/convert.test.js
@@ -50,15 +50,18 @@ describe('convert', () => {
   });
 
   it('bytesToBase64URL', () => {
+    const spy = jest.spyOn(global.console, 'error');
+
     expect(bytesToBase64URL(1)).toBe('AA');
     expect(bytesToBase64URL(2)).toBe('AAA');
     expect(bytesToBase64URL(3)).toBe('AAAA');
     expect(bytesToBase64URL([1])).toBe('AQ');
     expect(bytesToBase64URL([1, 0, 1])).toBe('AQAB');
     expect(bytesToBase64URL([0, 1, 1, 0])).toBe('AAEBAA');
-    expect(() => { bytesToBase64URL(''); }).toThrow(TypeError);
-    expect(() => { bytesToBase64URL('1'); }).toThrow(TypeError);
-    expect(() => { bytesToBase64URL('test'); }).toThrow(TypeError);
+
+    expect(spy).not.toHaveBeenCalled();
+    bytesToBase64URL('test');
+    expect(spy).toHaveBeenCalled();
   });
 
   it('base64URLToBytes', () => {
--- a/test/vendor/convert.test.js
+++ b/test/vendor/convert.test.js
@@ -50,8 +50,15 @@ describe('convert', () => {
   });
 
   it('bytesToBase64URL', () => {
-    expect(bytesToBase64URL('0000001')).toBe('AA');
-    expect(bytesToBase64URL('0000002')).toBe('AAA');
+    expect(bytesToBase64URL(1)).toBe('AA');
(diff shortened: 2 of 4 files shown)


test/vendor/utils.test.js
```diff
@@ -146,4 +157,88 @@ describe('utils', () => {
...
+    expect(first(['11', '22', '33'])).toBe('11');
+    expect(first('')).toBe(null);
+  });
+
+  it('last', () => {
+    expect(last('west')).toBe('t');
+    expect(last('foobar')).toBe('r');
+    expect(last(['11', '22', '33'])).toBe('33');
+    expect(last('')).toBe(null);
+  });
```
klahnakoski: `last` should treat strings the same as `first`

test/vendor/utils.test.js
```diff
@@ -146,4 +157,88 @@ describe('utils', () => {
...
+  it('zip', () => {
+    expect(zip([1, 2], [1, 2])).toStrictEqual([[1, 1], [2, 2]]);
+    expect(zip([1, 2, 3], [1, 2])).toStrictEqual([[1, 1], [2, 2], [3, undefined]]);
+  });
+
+  it('isMany', () => {
+    expect(isMany([])).toBe(true);
+    expect(isMany('')).toBe(false);
+    expect(isMany(null)).toBe(false);
+    expect(isMany({})).toBe(false);
```
klahnakoski: add a string to this

test/vendor/utils.test.js
```diff
@@ -146,4 +157,88 @@ describe('utils', () => {
...
+  });
+
+  it('reverse', () => {
+    expect(reverse([3, 2, 1])).toStrictEqual([1, 2, 3]);
+    expect(reverse([1])).toStrictEqual([1]);
+    expect(reverse(['t', 's', 'e', 'j'])).toStrictEqual(['j', 'e', 's', 't']);
```
klahnakoski: add a string to this

klahnakoski (approved):
This is good work!  You identified a number of oddities and problems that must be fixed! 

If you want to fix the code I would appreciate it.  If do not want to fix the code, that is ok:  I will add to this PR and fix the code before it is merged.  Your attribution will stay in the history.

Thanks again!

wicked-kiD:
> This is good work! You identified a number of oddities and problems that must be fixed!
> 
> If you want to fix the code I would appreciate it. If do not want to fix the code, that is ok: I will add to this PR and fix the code before it is merged. Your attribution will stay in the history.
> 
> Thanks again!

wicked-kiD:
> This is good work! You identified a number of oddities and problems that must be fixed!
> 
> If you want to fix the code I would appreciate it. If do not want to fix the code, that is ok: I will add to this PR and fix the code before it is merged. Your attribution will stay in the history.
> 
> Thanks again!

Hi. I'm going to fix it. I believe it will be some things to discuss. Thank you for your review and comments!

test/vendor/utils.test.js
wicked-kiD: In current implementation function is returning error if we'll pass string here. I can fix it, but I need some clarification about expected behaviour. Should it reverse arrays and strings?

test/vendor/convert.test.js
```diff
@@ -37,4 +43,19 @@ describe('convert', () => {
...
   });
+
+  it('escapeRegEx', () => {
+    expect(escapeRegEx('q2W#e.[{]}')).toBe('q2W#e\\.\\[\\{\\]\\}');
+    expect(escapeRegEx('./:loiep*&')).toBe('\\./:loiep\\*&');
+  });
+
+  it('bytesToBase64URL', () => {
+    expect(bytesToBase64URL('0000001')).toBe('AA');
+    expect(bytesToBase64URL('0000002')).toBe('AAA');
```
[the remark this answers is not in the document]
wicked-kiD: I'm not pretty sure how to test this method, will be grateful for any idea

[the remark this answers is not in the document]
klahnakoski: I would expect `0000001` to represent the byte sequence 0x30303030303031 (7bytes), which is then converted to base64 (10 bytes).  Then again, this is Javascript: If we pass it something that is not bytes, then we cannot really say what will come out.

Looking a the code:

    window.btoa(String.fromCharCode(...Array.from(new Uint8Array(bytes))))

some experimentation is required to understand how each function processes `0000001`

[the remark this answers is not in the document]
klahnakoski: ...or we can attempt to detect not-bytes, and throw an error

test/vendor/utils.test.js
klahnakoski: @wicked-kiD   The code assumes strings are atoms, not sequences of characters.  With this assumption, arrays will be reversed but strings are not

    reverse("hello") === "hello"

look at `toArray(value)` for a more specific definition.

klahnakoski: ...although

    reverse("hello") => ["hello"]

may be better.

klahnakoski: throwing an error is also legitimate; add the test that verifies the error is thrown.

test/vendor/utils.test.js
```diff
@@ -146,4 +157,88 @@ describe('utils', () => {
     expect(missing(0)).toBe(false);
     expect(missing({})).toBe(false);
   });
+
+  it('first', () => {
+    expect(first('test')).toBe('t');
+    expect(first('jest')).toBe('j');
+    expect(first(['11', '22', '33'])).toBe('11');
+    expect(first('')).toBe(null);
+  });
```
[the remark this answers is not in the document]
wicked-kiD: Sorry for some delay. I'm tried to understand the best way how to fix it, but I'm not sure. Can you share your vision about this cases:
1. last('') should be null or ''?
2. do we need handle type of the params to be sure that they are string or array? For example throw the error?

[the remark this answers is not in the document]
wicked-kiD: Also I'm thinking why don't you use Lodash here to do this operations? Is that because Lodash will _.head('test') === 't'? If you want, I can try to refactor all this utils.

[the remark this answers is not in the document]
klahnakoski: @wicked-kiD 

1. `last('')===null`  since the empty string is treated as `missing`
2. You may need to check if the parameter types to return the proper value; or you can leverage other functions to do type-check for you:  (inefficient, but works)

```
    last = x => {
        const list = toArray(x);
        return list[list.length-1]
    }
```

In general, this library is simplifying data manipulation by establishing a set of "equivalence classes" on data: Different data structures that represent the same thing should be equivalent.

Javascript has no operator for equivalent, so I will use `<=>`

The `missing()` function defines all the values that are equivalent to `null`:

    null <=> '' <=> [] <=> Number.POSITIVE_INFINITY  ...

This library also assumes everything is a list:

    null <=> []
    x <=> [x]
    "hello" <=> ["hello"]

The expectation is this library will provide the left-most as the canonical representative of the class.

> why don't you use Lodash here to do this operations


This library started because the previous owner did not like the size of lodash.   You are welcome to replace part of this library with lodash and see what happens to the compiled size and how it impacts the other code.  You are right there may be opportunities there.

test/vendor/convert.test.js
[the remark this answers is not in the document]
wicked-kiD: I'm afraid i still have a one reservation about that.

`window.btoa(String.fromCharCode(...Array.from(new Uint8Array(bytes))))
    .replace(/[+/=]/g, m => toURL[m]);`

we can replace with

`window.btoa(String.fromCharCode(...new Uint8Array(bytes)))
    .replace(/[+/=]/g, m => toURL[m]);`

Step by step we are having:
1. array filled with 0 with length of the bytes passed into the function (new Uint8Array(bytes))
2. string with spaces with the same length as the array (String.fromCharCode(...new Uint8Array(bytes)))
3. then we using window.btoa with that string and will have the string with only _A_ symbols and sometimes with _=_ that will be replaced at the last step.

In result I can't understand how can we use this function, seems it is working incorrect. What do you think?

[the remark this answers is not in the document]
klahnakoski: For simplicity, we can simply throw an error if a string is sent to `bytesToBase64URL`. I thought the string could be interpreted as bytes, but that is looking complicated.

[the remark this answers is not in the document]
wicked-kiD: I updated the PR, please take a look.

src/vendor/convert.js
```diff
@@ -107,7 +107,10 @@ const toURL = { '+': '-', '/': '_', '=': '' };
 convert from bytes to base64 (encoded to be URL safe)
  */
 function bytesToBase64URL(bytes) {
-  return window.btoa(String.fromCharCode(...Array.from(new Uint8Array(bytes))))
+  if (isString(bytes)) {
+    return null.f();
```
klahnakoski: This is not clear.  You can

    raise new TypeError("with some explanation")

or you can use the `Log` library

    raise Log.error("expecting array of ints, not {{type}}", {type: typeof bytes})

klahnakoski (changes requested):

src/vendor/convert.js
wicked-kiD: It need one more dependency into the package to throw new TypeError (@babel/plugin-proposal-throw-expressions).
For logging the error it will need to disable no-console rule for the line of code.

Which option would be better?

klahnakoski: Sorry for the delay.  I see I used Python's `raise` instead of ES6 `throw`.

Use 

    Log.error("expecting array of ints, not {{type}}", {type: typeof bytes})

which will not need the dependency, it will log the problem, and throw and error.

kimberlythegeek: Hi @wicked-kiD , I'm unsure why you marked this as resolved. I could be mistaken, but it doesn't appear you've made the changes that Kyle requested

wicked-kiD: > Hi @wicked-kiD , I'm unsure why you marked this as resolved. I could be mistaken, but it doesn't appear you've made the changes that Kyle requested

Hi @kimberlythegeek ! I did this changes, you can see them at 111-113 lines at convert.js. They are looking a bit different because example was not on JS. Also I added test case for this at lines 53, 62-64 at convert.test.js

Issue #665, opened 2020-03-06
Change from coveralls to codecov.io


klahnakoski:
Coveralls seems broken, and no longer can be used to sort  https://github.com/lemurheavy/coveralls-public/issues/1365

Change the `.travis.yml` file to use codecov.io instead

js636f:
@klahnakoski may I work on this issue?

kimberlythegeek:
Yes, thanks @js636f ! I am the other mentor on this project, feel free to ask me questions as well

silvererudite:
hi @kimberlythegeek can you pls recommend me a good first bug as I dont see any unassigned ones now? I'm an outeachy applicant,also interested to contribute to the project

js636f:
Thank you, @kimberlythegeek !

kimberlythegeek:
Hi @silvererudite ! Yes, I apologize, I realize there are not many to choose from right now. Kyle and I are going to begin filing more bugs today. You can also look at Perfherder-related bugs. Perfherder is another react dashboard that is outside the scope of the internship, but there are many more issues to choose from!

Perfherder is part of Treeherder: https://treeherder.mozilla.org/perf.html

[Perfherder bugs](https://bugzilla.mozilla.org/buglist.cgi?product=Tree%20Management&component=Perfherder&resolution=---&list_id=15132483)

kimberlythegeek:
@js636f You are quite welcome 🙂

kingsham:
> Yes, thanks @js636f ! I am the other mentor on this project, feel free to ask me questions as well

I need help contributing to a project

kimberlythegeek:
> I need help contributing to a project
Hi there--the best way to contribute is to look for unassigned bugs or issue's on a project's contribution page. We don't have many open issues for health.graphics (this project), but you can also contribute to [Perfherder](https://treeherder.mozilla.org/perf.html), another react dashboard. There are a lot more bugs to choose from. Simply comment and ask to be assigned. 

If you find a bug on bugzilla, please :needinfo me by check the box at the bottom of the page that says "Request information from" and enter my email: ksereduck@mozilla.com. I will receive an email notification and be able to respond faster

[unassigned Perfherder bugs](https://bugzilla.mozilla.org/buglist.cgi?cmdtype=runnamed&namedcmd=Perfherder%20Unassigned&list_id=15140577)

klahnakoski:
I think this is done

PR #666, opened 2020-03-06
Use codecov instead of coveralls

js636f:


kimberlythegeek:
#665

js636f:
@kimberlythegeek , I have a question. 
I'm on Windows 10 now and it seems to me it isn't fully suitable for this work. For example it gives this message when installing dependencies (and a lot of warnings):
`info fsevents@1.2.11: The platform "win32" is incompatible with this module.`
When I run tests it gives me a lot of messages like this (but all tests age green):
```
 console.log src/vendor/logs.js:207
      Cannot read property 'map' of undefined
      stacktrace
```
Should I use something like Ubuntu instead of Windows to contribute to this project?
Thanks!

kimberlythegeek:
@js636f You should still be able to run it on Windows. If you have a linux machine already set up, I would suggest using that, but it isn't necessary. 

Did a search for the error message you posted and found this: https://stackoverflow.com/questions/46929196/how-to-solve-npm-install-throwing-fsevents-warning-on-non-mac-os#47071650

Does that help?

js636f:
@kimberlythegeek , Thank you! It seems to me everything works on Windows, but I would like it to work without warnings. I tried `yarn install --ignore-optional`, but the messages are still here (I don't think ignoring warnings is a great idea, but as far as I understand that package isn't necessary in Windows). I'll try something else (npm, I think) and perhaps will finally move to Linux. I'll write if I find something interesting about these issues.

kimberlythegeek:
@js636f sorry to hear you’re having difficulties. It’s not easy for me to help debug because I’m on a different operating system, but I will do my best to help

js636f:
@kimberlythegeek Thank you! I think everything will be fixed soon because these difficulties are not too serious. In fact they are not prevent me from working. They are just a little troubles which can be fixed over time.

klahnakoski:
@js636f  I work on Windows, so I can help if that is still a problem.  Those warnings you mentioned are normal.  Everything still works.  It would be nice to get rid of the warnings, maybe we do not need the `fsevents` library?

The `Cannot read property 'map' of undefined` are warnings, not errors.  They are caused by the fact there is no node network library to make actual network calls during testing.  We should not have real network calls during testing; the intent is to mock those calls, eventually.

js636f:
@klahnakoski , Thank you very much! As I figured out we don't need fsevents on Windows (it's for Mac). I highly appreciate your explanation about those warnings. Now I much more confident in my working setup. Thank you again!

js636f:
@klahnakoski , Thank you! I'm happy to help.

Issue #675, opened 2020-03-19
New raptor-browsertime tests instead of old raptor-webext


klahnakoski:
raptor-webext tests for Fenix (Firefox Preview) will be switched off from March 26th now that @gmierz has enabled the raptor-browsertime tests.

klahnakoski:
From sparky: 

The Fenix Browsertime tests were merged into the Fenix repo: https://github.com/mozilla-mobile/fenix/pull/9087
Today, they had their first run: https://treeherder.mozilla.org/#/jobs?repo=fenix&tier=1%2C2%2C3&revision=756b1fbadd7b5aca72167982c600dc540552be04

They are running well, some errors are visible but it's nothing problematic. It looks like we will start turning off Raptor Fenix tests on March 26th/27th.
Note that our "baseline" measures for these various pageload tests will likely change with the move to Browsertime and these changes are unrelated to the application.

Visual metrics will also be running shortly, they are currently blocked on this PR which should land soon: https://github.com/mozilla-mobile/fenix/pull/9248

davehunt:
@airimovici are we ready for the switch from webext to browsertime tomorrow?

airimovici:
> @airimovici are we ready for the switch from webext to browsertime tomorrow?

Working on it. It will be ready for tomorrow

PR #677, opened 2020-03-25
Issue 675 - New raptor-browsertime tests instead of old raptor-webext

airimovici:
https://github.com/mozilla-frontend-infra/firefox-health-dashboard/issues/675


airimovici:
The data for some of the tests is missing, I have to double check the suite names for them
The measurement unit is different and also it acts as `lower-is-better` is `False`

**Previous version (raptor-webext)**

https://deploy-preview-677--firefox-health-dashboard.netlify.com/android/tp6m?test=cold-loadtime&browserPlatform=fenix-g5&past=month&ending=2020-03-25

raptor-webext-tp6m



**New version (raptor-browsertime)**

https://health.graphics/android/tp6m?test=cold-loadtime&browserPlatform=fenix-g5&past=month&ending=2020-03-25

raptor-browsertime-tp6m

airimovici:
The Measurement unit being displayed incorrectly and the `lowerIsBetter` being false is because the tests don't fit into the current condition that sets them when getting the data
           ```if (
              [
                'raptor-tp6',
                'raptor-firefox',
                'raptor-chrome',
                'raptor-google',
                'raptor-assorted-dom',
                'tp4',
                'tp5',
              ].some(prefix => suite.startsWith(prefix))
            ) {
              lowerIsBetter = true;
              unit = 'Duration';
            }```

airimovici:
@davehunt I think it is in a good state now and can be switched today if @klahnakoski is ok with the changes:

https://deploy-preview-677--firefox-health-dashboard.netlify.com/android/tp6m?test=cold-loadtime&browserPlatform=fenix-p2-aarch64&past=month&ending=2020-03-26

davehunt:
> @airimovici is it possible to retain the raptor-webext data for historical purposes? It looks like we're missing a couple of graphs. I see "23 of 28 sites reported". Also, there are no results for warm page loads, are we running these @gmierz? Finally, are we also running Speedometer via Browsertime now? If so, we'll need to migrate those graphs too.

Talking with @airimovici the missing results are due to [bug 1590763](https://bugzilla.mozilla.org/show_bug.cgi?id=1590763) and an issue with Google restaurants that we're going to file a bug for. We don't run warm page load tests against Fenix. Speedometer are run via the cron, which will be migrated separately.

PR #679, opened 2020-04-03
Show results for speedometer browsertime on fenix

airimovici:


airimovici:
These are the graphs for speedometer tests (I hid `geckoview` and `firefox ESR`):

speedometer-fenix

There are a few days when we got faulty results, but with this patch they were solved: https://phabricator.services.mozilla.com/D70535

PR #682, opened 2020-04-14
Issue 640 - Create dashboard page comparing Fission vs non-Fission perf test results

airimovici:


src/vendor/perfherder.js
```diff
@@ -129,6 +129,7 @@ const getFramework = async ({ repo, framework }) => {
             id: meta.id,
             framework: meta.framework_id,
             repo,
+            fission: Boolean(meta.extra_options) && meta.extra_options.includes('fission'),
```
klahnakoski, line 132: you can remove this and write `{eq: {extraOptions: "fission"}}` in your filters

airimovici: I does not work if I use it like that.

I'm not sure if that is because the `eq` [expression](https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/master/src/vendor/jx/expressions.js#L177) does not support it like that.

But there is also another problem. I also have to filter tests by "not having the fission extra option" (this problem might  be solved by using the expression `not`)

klahnakoski: You can keep it like this, or you can look at why it is not working. 

Here is the code that should run if using `{eq: {extraOptions: "fission"}}`:

https://github.com/mozilla-frontend-infra/firefox-health-dashboard/blob/a1312c3d44cdb3a8a4115e10c1f64997aef01312/src/vendor/jx/expressions.js#L174

> "not having the fission extra option"

yes, `{"not": expr}` should work on that

PR #689, opened 2020-05-18
NO MERGE - How fast is Autoland

klahnakoski:
This is a test to see how fast the health dashboard is with autoland.  It is only a naive replacement of "mozilla-central" with "autoland".


klahnakoski:
As I suspected, there are a number of performance series not run on `autoland`:  Either they are run on a different platform, or they are simply not run.  Each chart must be reviewed.

PR #696, opened 2020-05-28
Upgrade Travis, dependencies & Neutrino final release

armenzg:


armenzg:
@airimovici  FYI

I could not reproduce why Travis started failing on master. Upgrading all the dependencies and Node versions got it to green.

Please let me know if you notice any issues. It looked good from my testing.

PR #701, opened 2020-06-10
Remove "-cold" from Browsertime test names

davehunt:
This follows the change from [bug 1630675](https://bugzilla.mozilla.org/show_bug.cgi?id=1630675), where we remove '-cold' from the test name and instead include it in the extraOptions.


davehunt:
We have GeckoView cold results appearing. Once we see Fenix we can go ahead and merge this.

davehunt:
We now have Fenix results in the deploy preview. 🎉

PR #703, opened 2020-06-11
Switch tp6m Fennec68 to Browsertime

davehunt:
This patch switches Fennec68 to the Browsertime results. We should have some of these available over the coming weekend.


davehunt:
Fennec comparisons are now visible in the deploy preview. https://deploy-preview-703--firefox-health-dashboard.netlify.app/android/tp6m?test=cold-loadtime&browserPlatform=fenix-g5&past=month

